PR #18, opened 2016-02-05
Restore compression to base template with CssAbsoluteFilter

lewiscollard:
We stopped using the compressor because it broke calc queries, but CssAbsoluteFilter does not; it merely normalises URLs. This is a no-op in our case, so it effectively becomes a concatenator, which has two advantages:

1) As CSS in the frontend build is split out into two parts (stuff from CSS and stuff from Vue components), this means one HTTP request to load CSS rather than two.

2) A new filename is generated whenever the CSS changes, which means site updates do not require a subsequent cache clear.


lewiscollard:
Merge conflicts sorted.

Number 1) no longer applies now that there is no CSS file coming from Vue components, but 2) is still valid.

danielsamuels:
Thanks for the PR!

PR #24, opened 2016-02-18
Normalise database names: server.json should use cookiecutter.package_name

lewiscollard:
Currently, the `server.json` in the template uses `cookiecutter.repo_name` rather than `cookiecutter.package_name`. This will lead to problems when `repo_name` and `package_name` differ (for example, if there is a dash in `repo_name`).

This fixes server.json to use `package_name`.


danielsamuels:
Thanks for the PR!

PR #25, opened 2016-02-18
Use project_name for media/static paths on server

lewiscollard:
When repo_name is not the same as package_name, this causes the path to which `collectstatic` will save files and the paths which `manage.py deploy` expects diverge.

This normalises them by always using `package_name` as the base name for all three directories that server-management creates in /var/www/.


danielsamuels:
Can you remove the XFrameOptions line please?

lewiscollard:
Yeah, I have no idea how that happened, but done!

danielsamuels:
Thanks!

PR #29, opened 2016-03-02
Remove Python Social Auth from Suit menu.

lewiscollard:
It is extremely rare that it will be used (debugging Social Auth is the only scenario I can imagine), and so it's just clutter on the main menu.


danielsamuels:
Thanks for the PR!

Issue #32, opened 2016-03-07
Fix project name reference in gulpfile


danielsamuels:
https://github.com/onespacemedia/project-template/blob/16185e294f229a777734d236f07f9dc8b22ed5d6/%7B%7Bcookiecutter.repo_name%7D%7D/gulpfile.js/tasks/browserSync.js#L8

PR #33, opened 2016-03-09
Use cookiecutter.package_name rather than cookiecutter.project_name in browserSync.js.

lewiscollard:
Fixes issue #32.


danielsamuels:
Thanks for the PR!

PR #35, opened 2016-03-10
Updated assets and templates to match our latest base we'd like to use

dan-gamble:
- Updated stylelint config
- Updated gitignore
- Updated templates
- Updated CSS files
- Changed Revue to Vuex


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
```diff
@@ -61,7 +61,7 @@
           {% block content_below %}{% endblock %}
         </div>
 
-        {% include 'base/_footer.html' %}
+        {% include 'footer/_footer.html' %}
```
danielsamuels: I'm not sure if this change is to use the footer app or not. If it is, you should probably delete the base/_footer file.

dan-gamble: Yea, i also need to wrap it in a cookie cutter conditional as well. If the footer app is used then it's the added, else the `base`. Would removing it be a thing to do in the post hook?

danielsamuels: Yup.

danielsamuels:
Merged in 4c4aad8994743ed3a618a0d3dfdb087d5f7fa106.

PR #46, opened 2016-08-04
Add favicons to frontend build.

lewiscollard:
Right now the favicons include in the header is relying on favicons living in a `favicons` folder in the root static directory. As nothing is copied here by the build system (and the `static` directory is `.gitignore`'d) it means you effectively can't keep favicons in your repo.

This corrects this by 1) assuming favicons will live in `assets/img/favicons` and will be deployed to  `STATIC/build/img/favicons` 2) adding `.ico` files to the `images` gulp task so that they are copied with all the other image file types.

### DIFF
--- a/{{cookiecutter.repo_name}}/gulpfile.js/config.json
+++ b/{{cookiecutter.repo_name}}/gulpfile.js/config.json
@@ -46,7 +46,7 @@
     "images": {
       "src": "img",
       "dest": "img",
-      "extensions": ["jpg", "png", "svg", "gif"]
+      "extensions": ["jpg", "png", "svg", "gif", "ico"]
     },
 
     "fonts": {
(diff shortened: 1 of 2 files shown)


dan-gamble:
The main problem with this is because the `.gitignore` is too greedy.

I tend to change the last line to: `*/static/build`

I don't necessarily thing the favicons should go in the `assets` folder because nothing needs access to them and they don't need to be processed. The `static` folder isn't necessarily a byproduct of the `assets` folder and build system (the `static/build`) folder is, so i think `favicons` still going in `/static/` by default is fine imo

lewiscollard:
> and they don't need to be processed

Would it not be beneficial for these icons to be passed through the minification process in `npm run build`?

dan-gamble:
The build system doesn't do any kind of image minification because it broke on Linux

lewiscollard:
Ah, my mistake, I misread the output of `npm run build`.

PR #61, opened 2016-11-02
Add ALLOWED_HOSTS to local settings to fix Django 1.10.3 and 1.8.16

lewiscollard:
ALLOWED_HOSTS is now required to be set even with DEBUG on:

https://www.djangoproject.com/weblog/2016/nov/01/security-releases/

This adds 'localhost' and '127.0.0.1' as sensible defaults for local development.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/local.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/local.py
@@ -24,6 +24,11 @@ NODE_MODULES_ROOT = os.path.expanduser(os.path.join("~/Workspace/{{cookiecutter.
 
 SITE_DOMAIN = "localhost:8000"
 
+ALLOWED_HOSTS = ALLOWED_HOSTS + [
+    "127.0.0.1",
+    "localhost",
+]
+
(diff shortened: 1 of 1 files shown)


danielsamuels:
From that link:

> For convenience, if ALLOWED_HOSTS is empty and DEBUG=True, the following variations of localhost are allowed ['localhost', '127.0.0.1', '::1']. If your local settings file has your production ALLOWED_HOSTS value, you must now omit it to get those fallback values.

I'd propose setting the value to be `ALLOWED_HOSTS = []`

lewiscollard:
As I just found out (see above commit), compatibility with browsersync requires 0.0.0.0 to be in ALLOWED_HOSTS too, so emptying it won't work.

danielsamuels:
Oh, that's annoying. I'd probably still make it not extend the base `ALLOWED_HOSTS` though and just keep it as the local IPs.

lewiscollard:
Good idea. I'll do that and squash these commits.

danielsamuels:
I can squash on merge, so don't worry about that.

lewiscollard:
Oh yeah, forgot about that :)

danielsamuels:
Thanks for the PR!

Issue #54, opened 2016-10-26
JS to open all external links in a new window.


dan-gamble:
See any project I've built recently (usually 'external-links.js' or 'external-links/index.js'. Latest version is in the resources repo (LC)

danielsamuels:
Put into the base template.

PR #64, opened 2016-11-04
Add JS to open all external links in a new window. Ref #54.

lewiscollard:
I think we all agreed this is a reasonable idea!

Explanation: All links that do not point to the same domain are considered external.

'Same domain' is defined as 'same domain, when both are stripped of a leading `www.` string if present'.

As a bonus, it will also open social sharing links (Facebook, Twitter, LinkedIn) in a 600x300 popup (on devices where this is meaningful).

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/external-links/index.js
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/external-links/index.js
@@ -41,12 +41,12 @@ export default function () {
         // Add noopener and 'noreferrer' to work around this:
         // https://dev.to/ben/the-targetblank-vulnerability-by-example
         let rel = link.getAttribute('rel')
-        
+
         if (!rel) {
           // will be null if it is not set
           rel = ''
         }
-        
+
         link.setAttribute('rel', `${rel} noopener noreferrer`)
       }
     }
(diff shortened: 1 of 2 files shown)


dan-gamble:
Made a tiny syntax tweak to the `index.js`, but looks good to me 👍

danielsamuels:
I wonder if we could add `rel="nofollow"` to these links at the same?

lewiscollard:
Good question.

1) I know Google executes JS to some extent but I do not know if it will obey the attribute if it is set by Javascript.

2) Is it a good idea to nofollow anyway? If an external site is being linked to it probably has value and, to contrive a word to avoid a double-negative, deserves a `yesfollow` link. (This obviously isn't the case with user-supplied, rather than admin-supplied, external links, but these are relatively rare in our projects.)

danielsamuels:
For point 2, the general consensus is that you should nofollow all external links.

lewiscollard:
On point 1, at least as of 2015, it doesn't work:

> The nofollow in source code worked as expected (the link wasn’t followed). The nofollow in the DOM did not work (the link was followed, and the page indexed). Why? Because the modification of the a href element in the DOM happened too late: Google already crawled the link and queued the URL before it executed the JavaScript function that adds the rel="nofollow" tag. However, if the entire a href element with nofollow is inserted in the DOM, the nofollow is seen at the same time as the link (and its URL) and is therefore respected.

http://searchengineland.com/tested-googlebot-crawls-javascript-heres-learned-220157

danielsamuels:
Closing in favour of #67.

PR #65, opened 2016-11-15
Add base WYSIWYG title styles and add the options to the base WYSIWYG options. Resolves #51.

### LINKED ISSUE #51: WYSIWYG title classes in base settings.
## Lewis Collard

> WYSIWYG title classes in base settings.

Good idea. Plus stub classes in the WYSIWYG CSS?

WYSIWYG editor idea: can we make the default background colour in the admin white?

danielsamuels:


lewiscollard:
+1 from here.

danielsamuels:
@dan-gamble: Would you want those commented out `@apply`s in the template?

dan-gamble:
I think the commented out things are actually okay to leave, otherwise people can easily skim over them classes and think they've been added fine. Might even be worth adding more to it to let people know they still need to be updated

danielsamuels:
Leaving it as-is for now, we can always update the defaults in future.

PR #67, opened 2016-11-15
Add external links library. Resolves #54.

### LINKED ISSUE #54: JS to open all external links in a new window.

danielsamuels:


danielsamuels:
@dan-gamble: Could you give this a once over for me please?

lewiscollard:
I think `utils.js` does not give much away by its file name, and a single file for all utility things could both get uncomfortably large and have too much scope.

I think the original variant of this PR, with one top-level function per file in a `utils` directory, would be more sustainable.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/utils/external-links/index.js
```diff
@@ -0,0 +1,63 @@
+/*
+External links (defined as all links pointing to a domain that is not the
+same as the one from which the current document is being served)
+will have their 'target' attribute set to '_blank'.
+
+Social media sharing links will show in a popup.
+*/
+export default function () {
+  const links = [].slice.call(document.getElementsByTagName('a'))
```
dan-gamble: I'd change to to `Array.from(document.getElementsByTagName('a'))` as this is the natural API way of doing this.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/utils/external-links/index.js
```diff
@@ -0,0 +1,63 @@
...
+      if (domain !== thisDomain) {
+        link.setAttribute('target', '_blank')
+        // Add noopener and 'noreferrer' to work around this:
+        // https://dev.to/ben/the-targetblank-vulnerability-by-example
+        let rel = link.getAttribute('rel')
+        if (!rel) {
+          // will be null if it is not set
+          rel = ''
+        }
+        link.setAttribute('rel', `${rel} noopener noreferrer`)
```
dan-gamble: I always tend to have a white space before and after `if () {}` `for () {}` just for legibility.

dan-gamble (changes requested):

dan-gamble:
Not 100% sure if i did that right but i tried

dan-gamble:
I think a single `utils.js` file is fine for now we have very few functions we actually use as a `util` function, so at the current time i think a singular file is best. If this changes in the future and we start using JS more heavily and we build up a bigger library of utilities it would naturally progress into a folder with multiple files.

lewiscollard:
> I think a single utils.js file is fine for now we have very few functions we actually use as a util function, so at the current time i think a singular file is best

Reasonable. My opinion should not block this PR - it is a very weak preference :)

PR #69, opened 2016-11-16
Use single quotes consistently in settings.

lewiscollard:
The current settings files uses a mix of single and double quotes. One of them should be used consistently, and single quotes are as good as any (and is also consistent with our JS and CSS standards). This amends the settings to use the latter consistently.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/local.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/local.py
@@ -1,6 +1,6 @@
-"""
+'''
 Settings for local development.
-"""
+'''
 
 import os
 import os.path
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/local.py
```diff
@@ -1,6 +1,6 @@
-"""
+'''
 Settings for local development.
-"""
+'''
```
danielsamuels: I'd probably just remove this comment now, it doesn't serve much of a purpose, same goes for the one at the top of `base.py`.

lewiscollard: And the one in base.py was inaccurate! Done.

PR #72, opened 2016-11-16
Use single quotes consistently, everywhere.

lewiscollard:
Following on from #69. Use single quotes everywhere to replace the mixture of double and single quotes.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
@@ -1,5 +1,3 @@
-'''URL config for {{cookiecutter.repo_name}} project.'''
-
 from cms.forms import CMSPasswordChangeForm
 from cms.sitemaps import registered_sitemaps
 from cms.views import TextTemplateView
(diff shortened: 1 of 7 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
```diff
@@ -1,4 +1,4 @@
-"""URL config for {{cookiecutter.repo_name}} project."""
+'''URL config for {{cookiecutter.repo_name}} project.'''
```
danielsamuels: Can probably drop this

lewiscollard: I was thinking the same.

Can the same be done for `wsgi.py` too? That and the NewRelic archaeological find..

danielsamuels: Sure

lewiscollard: Done :)

PR #74, opened 2016-11-16
Add `get_[next,previous]_by_field` templatetags. Resolves #53.

### LINKED ISSUE #53: Move previous / next objects onto the model instead of the view
```
def next_career(self):
    try:
        return self.get_next_by_date_posted()
    except Career.DoesNotExist:
        return Career.objects.order_by('-date_posted', '-pk').last()

def prev_career(self):
    try:
        return self.get_previous_by_date_posted()
    except Career.DoesNotExist:
        return Career.objects.order_by('-date_posted', '-pk').first()
```

danielsamuels:


danielsamuels:
I'd be interested in getting some feedback on the implementation of this.

danielsamuels:
Sure, it's effectively a shortcut to having a method which invokes `get_next_by_date` on every model.

dan-gamble:
Yea. You might be able to expand on it a bit more and allow it to use any arbitary field.

Like you could check what the field's type is. If it's `DateTime` then you can use the inbuilt method if you not you could do something along the lines of https://github.com/onespacemedia/lcfi/blob/develop/lcfi/apps/programmes/models.py#L101 so it could take any field that is integer based and make it work for that as well.

Issue #59, opened 2016-10-26
Resurrect the Intrinsic mixin.


dan-gamble:
Mostly because it saves typing this a dozen times:

```
.whatever {
  position: relative;

  &::before {
    content: '';


    display: block;
    padding-bottom: percentage(9, 16);
  }
}
```

lewiscollard:
With this one approved: As we are going to have very few mixins in the base, and only for things that take parameters (this might even end up being the only one), should we just have a single `helpers/mixins.css` rather than resurrecting the `mixins` directory?

danielsamuels:
Where can I get this from and where's the best place to put it?

dan-gamble:
https://github.com/onespacemedia/owlstone-medical/blob/develop/owlstone/assets/css/helpers/mixins.css#L167

And i'd say this is the best place for it too a solo `mixins.css`

dan-gamble:
Akin to `utils.js`

lewiscollard:
On top of the above you'll need this:

https://github.com/onespacemedia/owlstone-medical/commit/a107a9c7b8a8549a04e637147b1d4a7338f2122c

PR #91, opened 2016-11-28
Add button mixins and some more WYSIWYG base

dan-gamble:
Adds the button mixins back
Fleshes out the WYSIWYG CSS file with more things we use on every project


danielsamuels:
How comes we're reverting this again?

dan-gamble:
Same reason as what's in https://github.com/onespacemedia/project-template/issues/59

Basically you can't do hovers, pseudo's etc. with just the `@apply` stuff so having a mixin for these makes styling them infinitely easier.

danielsamuels:
Why do they need to be `@apply` or `@Include` at all? Can't we just have a `.btn-Primary` class like we did before and put all the styles in there, and then use those whenever we need a button?

dan-gamble:
I started doing this then find we can't really use a single class for all use cases. For example when a button is on a dark background on desktop but then it ends up on a light background on mobile. I seem to have this every other project so i stopped trying to pin them all into the same class as then you have it done inconsistently. With gzip the `@include` way will barely have a size difference so i just prefer them to all be consistent.

PR #98, opened 2016-12-01
A bunch of optimizations to bundle sizes

dan-gamble:
Found these when starting NW Brown Portal project


danielsamuels (approved):

lewiscollard:
Am I reading it right in that this only loads Vue and the Promise polyfill when they are needed?

dan-gamble:
Nah, they are always loaded. Just in a separate file. You have just made me realise that it might be better to split this into 2 compressor things as well. The point of the Vendor split is that things rarely get added to this so we don't need to break the cache on update. But we compress them into 1 file at the moment.

PR #106, opened 2016-12-16
Add `pre-push` requirement script

danielsamuels:


dan-gamble (approved):

jamesfoley (approved):

lewiscollard (approved):

lewiscollard:
As curl is running silently here, I think it would be helpful to print a message informing the executor that pre-push hook is being installed.

I don't have any real objections (hence the approval!), but are there any potential failure cases around downloading and automatically installing whatever the latest version of `pre-push` is? I suppose not if `pre-push` is not going to change substantially.

danielsamuels:
Added some echos ref the above.

lewiscollard:
Spot on \o/

lewiscollard:
Hm. Just one more thing. What happens if you run `manage.py` from a directory that isn't the root directory of the project? I generally tend to sit in `<project-root>/<project-name>`...

danielsamuels:
It will likely fail because the `.git` folder doesn't exist.

danielsamuels:
```
$ ../manage.py runserver
chmod: .git/hooks/pre-push: No such file or directory
chmod: .git/hooks/pre-push: No such file or directory
Performing system checks...

System check identified no issues (0 silenced).
December 16, 2016 - 15:01:21
Django version 1.8.16, using settings 'imperial.settings.local'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```

dan-gamble:
Is there a git root variable you can use?

danielsamuels:
```git rev-parse --show-prefix```

Not sure how I'd implement that though.

**Comment by dan-gamble** [comment]
Can you not create a temporary variable with the output of that then do like `${temp}/.git/hooks/pre-push`?

**Comment by danielsamuels** [comment]
Ah:

```
git rev-parse --show-toplevel
```

dan-gamble:
This has been approved by all, will merge 👍

PR #107, opened 2016-12-22
Add safari iframe fix

dan-gamble:


lewiscollard (approved):

lewiscollard:
Excellent.

A future enhancement would be to see if we can do some kind of broken-iframe detection (maybe not possible), or maybe browser detection, just so browsers that do not break with Vue/Iframes don't have to load them twice. But this PR is an unquestionably good idea and is absolutely fine as-is.

danielsamuels:
Would it not be worth doing a UA check so it only applies to Safari? Seems to be quite a bit of repainting required for browsers which aren't breaking.

dan-gamble:
Good idea

danielsamuels (approved):

PR #101, opened 2016-12-03
Add babel-preset-env and babel-preset-latest

dan-gamble:
Updates some core babel plugins too
Also update autoprefixer's browser list

Fixes #83

PR #109, opened 2017-01-03
Move webpack bundles out of compressor

dan-gamble:
This lets us utilise lazy loading modules properly. When they are inside compressor all the lazy loaded bundles end up getting compressed into 1 so the JS errors when it tries to find one.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
@@ -90,9 +90,8 @@
       </div>
     {% endblock %}
 
-    {% compress js %}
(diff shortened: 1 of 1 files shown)


danielsamuels (approved):

lewiscollard:
Just double checking - built JS filenames are already randomised with their hash right? This looks to be the case but I do not remember the commit/PR that fixed this.

lewiscollard (approved):

dan-gamble:
I did that in a PR earlier: https://github.com/onespacemedia/project-template/pull/101/commits/efdf62b72f20c44925cc4e9921a62c7c66aaf1db

lewiscollard:
So you did! This all seems good to me.

PR #111, opened 2017-01-10
Add SECURE_PROXY_SSL_HEADER to settings.

lewiscollard:
As we are using HTTPS by default for all sites, and the latest server-management [adds the X-Forwarded-Proto header to the nginx config](https://github.com/onespacemedia/server-management/commit/47f96a94dd6281dc6d3088555d7d0eb7eedc957b), it makes sense to have this one-line change which makes `request.is_secure()` work (and consequently, canonical URLs and such use HTTPS as well).


danielsamuels:
I'd suggest putting this in production.py rather than base.py

lewiscollard:
Good idea, done!

danielsamuels (approved):

PR #116, opened 2017-01-25
Show all Save buttons in the admin

danielsamuels:
This file is a copy of [the Django suit file](https://github.com/darklow/django-suit/blob/develop/suit/templates/admin/submit_line.html), just with the button conditionals removed. This results in all of the save options appearing:


danielsamuels:
I'm just doing some checks to make sure they actually work if all included.

danielsamuels:
Looks like they all work.

dan-gamble (approved):

jamesfoley (approved):
Mixed feelings about the buttons always showing as you can turn them on per model admin.

But if needed you can just remove this file from the project if you wanted, so not much of a problem.

danielsamuels:
> Mixed feelings about the buttons always showing as you can turn them on per model admin.

The main issue is that if you turn on "Save as new", which we sometimes do to allow for duplicating items, you lose "Save and continue editing" and "Save and add another", which isn't ideal.

lewiscollard:
Could we move 'Save and add another' above 'Save as new'? I think it's possible that anyone that entered content a lot with the previous one could quite easily accidentally hit 'Save as new' by accident because it's in the same place that 'Save and add another' is.

lewiscollard (approved):

PR #124, opened 2017-02-27
Remove list style from ul by default

### LINKED ISSUE #118: Apply the Reset_List mixin to ul's by default
```postcss
@mixin Reset_List {
  margin: 0;
  padding: 0;

  list-style: none;
}
```

And then in the `wysiwyg.css` apply the browser's default to stop someone shipping something with no WYSIWYG list styles

dan-gamble:
Fixes #118


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/wysiwyg/wysiwyg.css
```diff
@@ -11,6 +11,14 @@
...
 /* We are styling root elements so we need access to type */
 /* stylelint-disable selector-no-type */
+p {
+  margin-bottom: 1vr;
+}
+
+ul {
+  list-style: disc;
+}
+
```
lewiscollard, line 21: As they're removed by the `@include Reset_List()`, would it be a good idea to restore browser default padding/margins?

dan-gamble: Whenever i see our designs they rarely have anything close to the browser default. It's only a default in Chrome as well tbf.

danielsamuels (approved):

PR #132, opened 2017-03-16
Order settings by their name.

lewiscollard:
Right now they come out in primary key order, which is not very obvious behaviour.


danielsamuels (approved):

danielsamuels:
Either add a migration, or update the base one.

lewiscollard:
As we're starting to do with the other apps, can we start off with an empty (minus `__init__.py`) migrations folder instead?

danielsamuels:
WFM

lewiscollard:
Tada!

danielsamuels:
Nice.

PR #140, opened 2017-05-08
[Fixes #117] Implement Typekit font caching

### LINKED ISSUE #117: Font cache ability
https://github.com/dan-gamble/TIL/blob/master/fonts/typekit-cache.md

I want to explore this a tad more.

Current i've been doing:

```javascript
  WebFont.load({
    typekit: {
      id: 'szt0stz'
    },
    timeout: 3000,

    active () {
      // Before the font downloads the underlying font is bigger so we get a bigger mark, if we
      // style it again when we get the new font flash we'll get the correct size
      stickyNav()
    }
  })
```

As their current async way doesn't have events like the sync version so you can do things on `active` for example like above.

@danielsamuels also mentioned doing:

```python
FONT_PROVIDER = 'typekit'
FONT_KIT_ID = 'ao3d87q'
FONT_TYPEKIT_CACHE = True`
```

This might also work for other font providers if they cache their font files like Typekit do

dan-gamble:


danielsamuels (commented):
I wonder if it's worth making it a little more generic and adding support for other font providers while we're here?

The change you could make would be to change `TYPEKIT_USED = True` to `FONT_PROVIDERS = ['typekit', 'fonts.com']`, and then have `TYPEKIT_KIT_ID = ''`, `FONTSCOM_KIT_ID = ''` etc.

Then you'd update the `_typekit.html` to be `_fonts.html` and use a conditional of `{% if 'typekit' in settings.FONT_PROVIDERS %}`.

dan-gamble:
I'm not sure how we can cache other providers. I think for the most part they cache themselves. I believe Typekit is the only one that needs manually caching our end.

danielsamuels:
It wasn't so much about the caching, more about the embed codes.

dan-gamble:
I'll have to refresh myself on how other providers give you fonts, i can look into that today hopefully.

danielsamuels:
I'm going to merge this in, we can make changes to it on future projects.

PR #142, opened 2017-05-08
[Fixes #134] Add a way to lazy load images in

### LINKED ISSUE #134: Handle images better
I think we could potentially come up with some kind of Vue component or JS snippet that emulates something like Medium.

https://thecoffeelicious.com/this-is-the-antidote-for-digital-narcissism-998b424a1ba1

You end up with HTML like:
```html
<figure name="4cb9" id="4cb9" class="graf graf--figure graf-after--p">
  <div class="aspectRatioPlaceholder is-locked" style="max-width: 550px; max-height: 496px;">
    <div class="aspectRatioPlaceholder-fill" style="padding-bottom: 90.2%;"></div>
    <div class="progressiveMedia js-progressiveMedia graf-image is-canvasLoaded is-imageLoaded" data-image-id="0*DE0XsDzUJvFtS95w.jpg" data-width="550" data-height="496" data-scroll="native">
      <img src="https://cdn-images-1.medium.com/freeze/max/30/0*DE0XsDzUJvFtS95w.jpg?q=20" crossorigin="anonymous" class="progressiveMedia-thumbnail js-progressiveMedia-thumbnail">
      <canvas class="progressiveMedia-canvas js-progressiveMedia-canvas" width="75" height="67"></canvas>
      <img class="progressiveMedia-image js-progressiveMedia-image" data-src="https://cdn-images-1.medium.com/max/800/0*DE0XsDzUJvFtS95w.jpg" src="https://cdn-images-1.medium.com/max/800/0*DE0XsDzUJvFtS95w.jpg">
      <noscript class="js-progressiveMedia-inner">&lt;img class="progressiveMedia-noscript js-progressiveMedia-inner" src="https://cdn-images-1.medium.com/max/800/0*DE0XsDzUJvFtS95w.jpg"&gt;</noscript>
    </div>
  </div>
</figure>
```

Few notes:

* They progressively load images
* They use the "Intrinsic" style to do images
* They don't blanket images with a singular aspect ratio, they use the uploaded sizes to do it

Advantages to this:

* Better page load times due to progressively loading images
* No layout shifting as they have the `aspectRatioPlaceholder-fill` to hold the size even when the image isn't loaded
* A much nicer overall experience.

I believe this could be achieved relatively easily with the `media` app change where it records the size on upload. We would also be able to supply aspect ratio props so we could ensure a certain ratio is followed on cards for example where everything needs to be uniformly styled.

Edit**
Link: https://jmperezperez.com/medium-image-progressive-loading-placeholder/

dan-gamble:


danielsamuels (approved):
It would be useful to have some documentation on how to use it, but everything looks OK to me.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/vue/components/images/LazyImage.css
```diff
@@ -0,0 +1,69 @@
...
+  width: 100%;
+  height: 100%;
+
+  opacity: 0;
+
+  transition: opacity 0.5s linear;
+
+  @supports (object-fit: cover) {
+    object-fit: cover;
+  }
```
lewiscollard, line 44: Is any of the object-fit stuff (and corresponding hacks to make it work in IE11) needed? By definition, it will always cover the full size of the container as it is absolutely positioned within it.

dan-gamble: The `object-fit` stuff is helpful, the `@supports` is probably not needed now tbf because of the `img-Image_Image-ie` below.

Even if the image is absolutely positioned it doesn't guarantee the image will be scaled properly, `object-fit: cover` basically make the image work like `background-size: cover` so if the image is actually bigger than the image container it's given it will pull itself off the edges to maintain it's size.

dan-gamble:
I added a small piece showing how to use it where the templatetag is defined, do you have a suggestion on a place where i could better document it?

danielsamuels:
A docstring on the templatetag sounds fine. I'm not seeing that commit though.

dan-gamble:
It's in the initial commit https://github.com/onespacemedia/project-template/pull/142/commits/d78e29560ed90ad5be4118c6fb08477b1d097a86#diff-1f46642d94019a5cf7ff90c3bcaf36d1R53

danielsamuels:
Ah cool, that works.

PR #152, opened 2017-06-13
Ensure that IE11 never renders in compatibility mode.

lewiscollard:
This avoids rare situations in which IE11 decides it is on an Intranet and that it is really Internet Explorer 8.


danielsamuels (approved):

danielsamuels:
Thanks for the PR!

Issue #148, opened 2017-05-17
Add eslint-plugin-compat for better browser compatibility checking according to our browserslist


dan-gamble:
https://github.com/amilajack/eslint-plugin-compat

Issue #146, opened 2017-05-10
Look in to prettier and something along the lines of a CSS equivalent.


dan-gamble:
[stylefmt](https://github.com/morishitter/stylefmt)
[perfectionist](https://github.com/ben-eb/perfectionist)
[prettier](https://github.com/prettier/prettier)

This means we could get a load of our stylistic rules out of the configs and use the linters for more important stuff. This should hopefully slightly speed up workflow and stop annoying error takeovers for missing random spaces.

dan-gamble:
Standard stuff:

https://github.com/sheerun/prettier-standard
https://github.com/dtinth/prettier-standard-formatter

PR #153, opened 2017-06-15
Fixes for #146, #148, #149

dan-gamble:
`stylelint --fix` is a tad spotty with our `property-definitions.css`. I've [raised an issue](https://github.com/stylelint/stylelint/issues/2643) so hopefully that's sorted soon and that could be added in.

Even though the `stylelint --fix` stuff is spotty i think this is fine to merge as we have the `prettier` part working fine 👍


danielsamuels:
What's the max line length set to? It seems pretty narrow.

dan-gamble:
[80](https://github.com/prettier/prettier#options) They recommend to not go over it. Baring in mind this doesn't impact how we write JS, just the way we read it i figured it was fine.

We can still write lines 150 long if we want, they will just get formatted down to 80 on commit.

danielsamuels:
It's just when I see lines like this:

```
    const currentPos =
       window.scrollY || window.pageYOffset || document.documentElement.scrollTop
```

It just seems like a bit of a pointless wrap, if anything I would say it's less readable than just having it on one line.  I see what they mean about the reasoning behind not increasing it, you definitely wouldn't want all lines at 100, 120 or w/e.  I dunno, I guess I just don't like the feeling that it makes _some_ code look worse.

dan-gamble:
If it's really causing some problems on certain lines you could just add `// prettier-ignore`. If it becomes a wider problem we could potentially look at other avenues.

danielsamuels (approved):

PR #154, opened 2017-06-15
Feature/npm packages update

dan-gamble:
This is reliant on #153 and can't be merged until that has.


{{cookiecutter.repo_name}}/package.json
```diff
@@ -198,7 +198,7 @@
         1
       ]
     }
-  },
+  }
```
lewiscollard: Is this missing a comma?

dan-gamble: Yep!

{{cookiecutter.repo_name}}/package.json
```diff
@@ -444,7 +444,6 @@
       "at-rule-name-case": "lower",
       "at-rule-name-space-after": "always",
       "at-rule-no-vendor-prefix": true,
-      "stylelint-disable-reason": "always-before",
```
lewiscollard, line 402: I personally rather liked this rule. Is there a newer alternative to it, or a reason it is deprecated?

dan-gamble: > The following rules did not work well.
stylelint-disable-reason could not enforce providing a reason

Is all they seem to say. I was hit or miss on it personally.

danielsamuels: I liked it, but there's nothing stopping you from continuing adding a reason above regardless of the enforcement.

dan-gamble: Yea, and if you still wanted to enforce it, stylelint rules are pretty easy to write. Could just take their source code and create a plugin from it

lewiscollard: Thinking about it, I see their logic completely (they they can only enforce that _there is a comment above it_, not an actual _reason_). I plan to always add a reason either way. :)

lewiscollard (approved):

PR #155, opened 2017-06-15
Update urls.py to be 1.11 compatible

dan-gamble:


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
```diff
@@ -16,30 +19,26 @@
...
     # Basic robots.txt.
     url(r'^robots.txt$', TextTemplateView.as_view(template_name='robots.txt')),
 
     # There's no favicon here!
     url(r'^favicon.ico$', generic.RedirectView.as_view(permanent=True)),
-] + static(
-    settings.MEDIA_URL, document_root=settings.MEDIA_ROOT
-) + static(
-    settings.NODE_MODULES_URL, document_root=settings.NODE_MODULES_ROOT
-)
```
lewiscollard: Do we not need these in 1.11?

dan-gamble: As far as i can see these have never been used. I know 100% the `node_modules` one hasn't

lewiscollard: I think the first is necessary for having media files in local development. I just removed it from a project and uploads stopped working.

lewiscollard: ^^ By that I meant 'files that have been uploaded in the admin no longer show on the front-end', not 'uploading doesn't work'. :)

dan-gamble: Ah gotcha, will add that back then 👍

danielsamuels (approved):

PR #156, opened 2017-06-21
Bump python env version in manage.py

dan-gamble:


{{cookiecutter.repo_name}}/manage.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7
+#!/usr/bin/env python3.6
```
danielsamuels: You can just use `python`

danielsamuels (approved):

PR #162, opened 2017-06-30
Fix concatenation of list and tuple in production settings.

lewiscollard:


danielsamuels (approved):

danielsamuels:
This seems to be causing issues in new builds now. base.py is using [a tuple](https://github.com/onespacemedia/project-template/blob/develop/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.package_name%7D%7D/settings/base.py#L195), but production.py (as of this PR) is using [a list](https://github.com/onespacemedia/project-template/blob/develop/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.package_name%7D%7D/settings/production.py#L19).

PR #164, opened 2017-07-07
Feature/app consolidation

dan-gamble:
I've also added a base Header and Footer models inside the `site` app as they don't really warrant their own apps but can be useful for housing language editable stuff.


danielsamuels (approved):
I'm happy with this, I think we should try and get the sections modal thing fixed as well if we haven't already. But we can do that separately.

PR #170, opened 2017-09-11
Slightly tweak media definitions

dan-gamble:


lewiscollard:
LGTM, but we'll need to change `mediaBreakpoints` in `assets/js/utils.js` too :)

dan-gamble:
Good shout, would love to find a way to sync them.

PR #171, opened 2017-09-11
Bump package.json versions

dan-gamble:


{{cookiecutter.repo_name}}/package.json
```diff
@@ -70,13 +70,13 @@
...
-    "prettier": "^1.4.4",
-    "prettier-standard": "^5.1.0",
+    "prettier": "^1.6.1",
+    "prettier-standard": "^6.0.0",
     "require-dir": "0.3.2",
     "sanitize.css": "^5.0.0",
     "standard": "^10.0.2",
     "style-loader": "^0.18.2",
-    "stylelint": "^7.11.0",
+    "stylelint": "^8.1.1",
```
lewiscollard, line 79: This will need to be done in circle.yml as well I think.

dan-gamble: +1

PR #175, opened 2017-09-11
[WIP] Stripped django-watson from all models

jamesfoley:


lewiscollard:
Discussed in dev lunch. We decided Watson is good enough for the majority of projects on which we work. If we need to use something bigger and better on large-scale projects, we will.

Issue #181, opened 2017-09-26
Fix incorrect stylelint disable lines


jamesfoley:
selector-max-type etc

lewiscollard:
Is `selector-max-type` incorrect?

jamesfoley:
`selector-no-type` and `selector-no-universal` are the old ones that need fixing to `selector-max-type` and `selector-max-universal`

lewiscollard:
Fixed in #182.

Issue #180, opened 2017-09-25
Remove LazyImage ref from Vue index.js


jamesfoley:

lewiscollard:
Fixed in #182.

PR #182, opened 2017-09-26
[#181, #180] A few frontend build fixes

lewiscollard:
* Update stylelint-disable lines to use `selector-max-type` and `selector-max-universal` rather than `selector-no-*`.
* Remove a reference to a dead Vue component.
* Force prettier to use single quotes on CSS - it was converting single quotes to double quotes which was causing pre-commits to fail.

### DIFF
--- a/{{cookiecutter.repo_name}}/package.json
+++ b/{{cookiecutter.repo_name}}/package.json
@@ -112,7 +112,8 @@
       ],
       "**/!(migrations|tests)/*.py": [
         "pylint --load-plugins pylint_django,pylint_mccabe -d missing-docstring,invalid-name,no-init,too-many-ancestors,no-member,line-too-long,attribute-defined-outside-init,too-few-public-methods,no-self-use,unused-argument,protected-access,locally-disabled,duplicate-code,fixme --reports=n",
-        "isort --skip-glob=.venv --skip-glob=node_modules --skip-glob=*/migrations/*"
+        "isort --skip-glob=.venv --skip-glob=node_modules --skip-glob=*/migrations/*",
+        "git add"
       ]
     }
   },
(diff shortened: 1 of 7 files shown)


{{cookiecutter.repo_name}}/package.json
```diff
@@ -106,7 +106,7 @@
         "git add"
       ],
       "**/!(definitions)/*.css": [
-        "prettier --write",
+        "prettier --write --single-quote",
```
dan-gamble: This should just be `prettier-standard --write` The same as the `*.js` one.

dan-gamble (changes requested):

{{cookiecutter.repo_name}}/package.json
lewiscollard: I thought `prettier-standard` was a Javascript-specific formatter.

dan-gamble: Nah, `prettier-standard` is just `prettier` but with all the rules of StandardJS (Like single quotes)

lewiscollard:
Now has a fix for #179.

dan-gamble:
You could add `git add` to https://github.com/onespacemedia/project-template/blob/4a4c36b23644a9e822aa153772b8afed7adf6c11/%7B%7Bcookiecutter.repo_name%7D%7D/package.json#L115 The Python portion of this. I know that's an outstanding bug

dan-gamble:
Looks okay to me though, could be worth starting a project with it and seeing if you can do `yarn run dev` from the get go

lewiscollard:
@dan-gamble,

> You could add git add to [the] Python portion of this. I know that's an outstanding bug

Done.

> Looks okay to me though, could be worth starting a project with it and seeing if you can do yarn run dev from the get go

Just done so with all of the commits in this PR. `yarn run dev`, `build`, `lint:css`, `lint:js` are all working. :)

dan-gamble (approved):

dan-gamble:
Awesome, if that's the case i'd say this is g2g 👍

PR #183, opened 2017-09-27
Exclude sitemap.xml and its friends from Jinja2 processing.

lewiscollard:
This fixes an error wherein Jinja2 complains about `{% spaceless %}` in the CMS sitemap template.


dan-gamble:
The regex is for folders, you can probably remove the `.xml` and it should work the same providing it's in a `/sitemap/` folder

lewiscollard:
Hm, the `.xml` was required (the template lives at `templates/sitemap.xml` in the app folder), but after thinking about your comment I realised part of the match was not relevant so I simplified it in 328633f.

dan-gamble:
That makes sense i figured it would just `/templates/sitemap/sitemap.xml`. If it's just in `/templates` then the `.xml` is needed 👍

danielsamuels (approved):

PR #188, opened 2017-10-03
Assorted improvements to the people app.

lewiscollard:
* Inherit Person model from SearchMetaBase, rather than PageBase (the latter has a 'Title' field, but it doesn't map semantically onto a person's formal title)
* Consequence of the above: Remove navigation fields from person model that didn't do anything.
* Fix `title` in the template context for people detail being incorrect
* Capitalise 'LinkedIn URL' and 'Twitter username' correctly
* Add team to `list_filter` on people
* Remove get_linkedin_url function (URL-ness is enforced by the default widget)
* Use Python 3 super() calls.
* Prefer dangly commas in people models
* Prefer `object` and `object_list` in `person_list.html`
* Fix admin error (`order` is removed from the fieldsets for sortables)


lewiscollard:
> I'd add the get_linkedin_url and get_twitter_url methods to the Person model that we have on random objects. They are really helpful.

`get_twitter_url` is already there :) URLness of LinkedIn URLs is enforced by it being a URLField.

> I'd maybe align linkedin_url and twitter_username they currently require different things. I know the method will not make that matter but for the user having both fields be the same type would be slightly easier to understand.

I somewhat agree. However, it is possible, even common, for LinkedIn users to not have a username. We could align these by requiring a URL, rather than a username, for Twitter, but it's also somewhat common to refer to people by their Twitter username (e.g. `Follow @lewiscollard`), and having a username permits integrations. I don't have particularly strong feelings on this, though.

dan-gamble:
Yea i had the same thought process and agree with you that Twitter == username, Linked == URL. Just wondered if we could neutralise our wording so that we didn't push towards either and they can just add what they want.

I liked the method you wrote that would allow them to do `twitterUsername`, `@twitterUsername`, `httpToTwitterUsername`.

So having that work on either field with neutral working would be my preference but as well i don't have strong feelings on either way. Was just an observation from the users perspective.

lewiscollard:
> Just wondered if we could neutralise our wording so that we didn't push towards either and they can just add what they want.

Come to think of it, this is a really good idea.

lewiscollard:
Alright, thanks to your feedback I've renamed these fields so that they can accept just about anything that looks something like a username or URL.

Twitter is now merely called `twitter` and accepts `onespacemedia`, `@onespacemedia`, `twitter.com/onespacemedia`, `https://twitter.com/onespacemedia`, etc, all of which are normalised to `onespacemedia` by the form.

LinkedIn can accept `danielsamuels`, `@danielsamuels`, `www.linkedin.com/in/danielsamuels`, etc, all of which are normalised to `https://www.linkedin.com/in/danielsamuels` by the form.

How's that? :)

dan-gamble:
Sounds perfect 👍

dan-gamble (approved):

PR #191, opened 2017-10-07
Admin based email system

danielsamuels:
Hello everyone, I present to you an evolution of the `utils/email.py` / `send_email_by_reference` system.  This is an admin based version of what we've been using for a while, which allows admins to define template contents and view the stream of outgoing mail.

Code-wise it integrates into your code in basically the same way, you import `send_email` and pass it a reference code and an email address at the minimum, with the ability to add kwargs where needed.

Email templates are written in Markdown and the HTML version is automatically generated. A basic template (ripped from CW, their logo persists in it) is included in this PR.

There's some fancy admin integrations too, so you can see how your emails will look and did look (for emails sent). Screenshots below.

I'm open to reviews, adoration, gifts etc.

**Managing a template:**

**Viewing the log of sent emails:**

**Viewing a previously sent email:**

### DIFF
--- a/{{cookiecutter.repo_name}}/doc/sending-emails.md
+++ b/{{cookiecutter.repo_name}}/doc/sending-emails.md
@@ -1,0 +1,46 @@
+Included in this project is a very simple to use, and simple to integrate email system. Designed for brochureware and large applications alike.  Email templates are defined in the admin system and are able to be called from anywhere within a Django project.
+
+## Requirements
+
+Only really tested with Python 3.6, Django 1.11 and PostgreSQL. You're welcome to try it with other versions, but it might break.  Python versions older than 3.6 probably won't work due to the use of f-strings.  Databases other than PostgreSQL probably won't work due to the use of `JSONField`.
+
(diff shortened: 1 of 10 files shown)


lewiscollard:
I think this is awesome. My only small suggestion is that there's some sort of guard against emails being deleted; this could make content break code. Maybe an 'Allow deletion' checkbox guarding the delete button so that deleting it requires a small amount of extra effort.

danielsamuels:
I've made some changes on Quotient to disable deleting template objects or changing the reference. Quite restrictive, but I don't think too unreasonable.  What do you think?

https://github.com/onespacemedia/quotient-sciences/blob/448d0c735c0e81627c60b81ffcf76f401763a2bc/quotient/apps/emails/admin.py#L10-L19

lewiscollard:
Not unreasonable at all, in lieu of `User.is_really_REALLY_superuser`.

danielsamuels:
I've used `if 'onespacemedia.com' in request.user.email` before..

lewiscollard:
That's not a bad idea either.

dan-gamble (approved):

lewiscollard (approved):

PR #194, opened 2017-10-13
Add a few nice admin improvements for the default apps

lewiscollard:
Per JD's latest:

* Add last modified and categories to news admin `list_display`
* Move news article categories out of "Publication"
* Add category to FAQs `list_display`
* Remove almost-never-used `content_primary` from news categories model
* Add job title and team to people admin.
* Add image to `list_display` of news articles, people and events. A consequence of this is that `utils/admin.py` now has a helper `HasImageAdminMixin` for any model that has an image (it assumes the field is called `image`, but this can be overriden) - inherit from it and add `get_image` to the `list_display` of your model admin to render the image in the list.


danielsamuels:
For the image listings, I would suggest the following (example taken from people):

```py
list_display = ['get_preview', '__str__', 'job_title', 'email', 'is_online']
list_display_links = ['get_preview', '__str__']
```

I personally feel like the layout works well (and matches media.File):

lewiscollard:
Good idea @danielsamuels, and done. :) I've kept the name "image" rather than "preview" though for the column, if that's alright - it seems to me that "preview" makes sense for the `media.File` admin, but less so for an article or a person.

danielsamuels:
Sure, it was just a copy / paste from what we're using on QS.

danielsamuels (approved):

PR #199, opened 2017-10-20
Lazy image object fit detection fix for Edge

lewiscollard:
This changes the way lazy image detects `object-fit` support. Previously, the Javascript was testing for it by seeing if `object-fit` was in the document style (which would not be the case on Edge <= 15), but then applied IE11-specific browser hack to show the fallback which Edge ignored.

This changes it to show the fallback in the default rule, then uses `@supports (object-fit: cover)` to hide the fallback.

This fixes lazy images in Edge <= 15 (about 2% usage globally).

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/modules/lazy-image.css
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/modules/lazy-image.css
@@ -51,7 +51,7 @@
   background-position: center;
   background-size: cover;
 
-  @supports (object-fits: cover) {
+  @supports (object-fit: cover) {
     display: none;
   }
 }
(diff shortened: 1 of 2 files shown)


dan-gamble:
This LGTM. Only thing i'll say is i'd potentially hold off merging this as i would like to get https://github.com/onespacemedia/the-bradfield-centre/commit/1cb777d9b84a72ae92d6b5712f82182f280e6b16 in as well which is `lazy-image` related.

I'd also like `lazy-image` to be truly `lazy` with `IntersectionObserver` so could potentially roll all these in to one.

Now i'm saying this, that list might be worth splitting in to 3 separate PR's.

dan-gamble (approved):

danielsamuels:
Going to merge this one, we can decide on the other changes in another PR.

Issue #99, opened 2016-12-01
Update password hashers to include Argon2


danielsamuels:
https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#using-argon2-with-django

Support was added in 1.10, so we can either backport it (potentially dangerous?) or wait until the 1.11 LTS release.

danielsamuels:
Waiting on onespacemedia/cms#126.

lewiscollard:
Fixed by #200

PR #200, opened 2017-10-23
[#99] Make Argon2 the preferred password hasher.

lewiscollard:
As #99 has been around for a while :) Tested on Bradfield, works.

As this is the project template which I don't think we have ever used to upgrade an existing installation, should I just remove all the other hashers? It doesn't actually *hurt* them being there of course, as Argon2 will be used for all new accounts.

### DIFF
--- a/{{cookiecutter.repo_name}}/requirements.txt
+++ b/{{cookiecutter.repo_name}}/requirements.txt
@@ -1,5 +1,4 @@
 argon2-cffi
-bcrypt
 django>=1.11,<1.12
 django-cachalot
 django-compressor
--- a/{{cookiecutter.repo_name}}/requirements.txt
+++ b/{{cookiecutter.repo_name}}/requirements.txt
@@ -1,3 +1,4 @@
+argon2-cffi
 bcrypt
 django>=1.11,<1.12
 django-cachalot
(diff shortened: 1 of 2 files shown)


danielsamuels:
I don't think it's necessary to remove the other hashers from the list, but it might be worth removing bcrypt as I believe that requires an expensive install both in PT and in SM deploy.

lewiscollard:
Good point, and removed.

It struck me that if we were upgrading a previous DB via the PT somehow (and I don't know that it would ever happen like that), all of those in at least the last two years would have been hashed by bcrypt, so the other password hashers were entirely redundant. So I've killed them off.

danielsamuels:
The only thing I'm slightly concerned about is that if for some reason Argon2 fails, there's not going to be anything to fall back on. Would this ever be the case though?

lewiscollard:
For generating new password hashes, it looks from the Django source (`contrib/auth/hashers.py`) that the other algorithms mentioned will never be used as a fallback - the other hashers are only used as candidates for checking existing passwords. The primary hasher failing would happen noisily regardless of any other methods in `PASSWORD_HASHERS`, which is a good thing.

danielsamuels (approved):
Ah, I misunderstood how the system worked. I'm happy with this then.

PR #205, opened 2017-10-24
Sections app: Change the default link field.

lewiscollard:
It's not always a 'button', so rename it appropriately, as we seem to do this in most projects early on :)

Also add a link page option for ease of use and handle PublicationManager throwing 404s on this, which comes up from time to time.


dan-gamble (changes requested):
`fk_name = 'page'` needs adding to `ContentSectionInline` otherwise it throws an error as it's not sure which `pages.Page` fk to use.

dan-gamble (approved):

PR #208, opened 2017-10-25
A bunch of improvements to the settings app

lewiscollard:
* Add an HTML setting type
* Make `Setting.value` a cached property as it shouldn't change over the lifetime of an instance
* Add get_setting_value helper function (from CW)
* Add a template tag to get a setting rather than its value
* Make defaults in template tag work
* Add an admin warning about changing the type of an existing setting
* Add unit tests


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/settings/models.py
```diff
@@ -53,10 +61,12 @@ class Meta:
     def __str__(self):
         return self.name
 
+    @cached_property
     def value(self):
         return {
             'string': self.string,
             'text': linebreaksbr(mark_safe(self.text)),
+            'html': mark_safe(self.html),
```
danielsamuels, line 70: It doesn't look like you've added 'html' as a choice in the `type` field.

danielsamuels (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/settings/models.py
lewiscollard: Thats because I'm an idiot. Fixed! :)

danielsamuels (approved):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/settings/static/settings/js/admin/fields.js
```diff
@@ -9,7 +9,8 @@ function hide_inline_section_fields () {
     '.field-string': ['string'],
     '.field-text': ['text'],
     '.field-number': ['number'],
-    '.field-image': ['image']
+    '.field-image': ['image'],
+    '.field-html': ['html'],
```
danielsamuels: lint:js hasn't picked up on this, but we don't usually have dangling commas in JS

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/settings/templatetags/settings.py
```diff
@@ -1,14 +1,18 @@
...
-def setting(key):
-    obj = get_setting_util(key=key, default=None)
-    if not obj:
+def setting(key, default=None):
+    value = get_setting_value(key=key, default=None)
+    if value is None:
+        # rather than 'return default or key' - means coverage isn't fooled?
+        if default is not None:
+            return default
         return key
```
danielsamuels: A minor point, but personally I've never seen the value in returning the settings key to the front-end. I know it can be useful to see what you need to add, but I feel like we should be ensuring they're entered. I don't feel like the risk of having "facebook-url" output on the frontend is worth it.

lewiscollard: I can't muster an opinion either way - I would not mind removing this at all if you think that's the right thing.

lewiscollard: Come to think of it: Given that we have a `default` parameter in case we really want to see SOMETHING on the front end - yeah, let's go for it.

PR #212, opened 2017-10-25
Fix lint and lint-staged paths

### LINKED ISSUE #197: Fix JS paths for linting + precommit
* `lint:js` checks `{{cookiecutter.package_name}}/assets/js/**/*.js`, which won't catch anything that sits directly in `assets/js` (e.g. `main.js`)
* `lint-staged` checks a different glob: `*.js`, which differs and will also catch changes on  `sections/templates/admin/pages/page/sections.js` (which causes a syntax error because it's a Django template)

dan-gamble:
Fixes #197

### DIFF
--- a/{{cookiecutter.repo_name}}/package.json
+++ b/{{cookiecutter.repo_name}}/package.json
@@ -93,26 +93,27 @@
     "dev": "gulp",
     "build": "gulp production",
     "lint:css": "stylelint \"{{cookiecutter.package_name}}/{assets/**,apps/*/assets/css}/*.css\"",
-    "lint:js": "eslint --ext .js,.vue {{cookiecutter.package_name}}/assets/js/**/*.js"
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repo_name}}/package.json
```diff
@@ -93,26 +93,27 @@
...
         "prettier-standard",
+        "eslint --fix",
         "git add"
       ],
       "*.vue": [
         "eslint --fix",
         "git add"
       ],
-      "**/!(definitions)/*.css": [
+      "{{cookiecutter.package_name}}/{assets/**,apps/*/assets/css}/*.css\"": [
```
lewiscollard, line 109: As this no longer excludes the definitions directory, does this depend on #213?

dan-gamble: Yep!

PR #213, opened 2017-10-25
Strip postcss-apply

dan-gamble:
There were a few reasons to this:

* By stripping postcss-apply we get to utilise stylelint --fix. That currently
gets broken in our property-definitions.css when we have postcss-apply.
* @apply has been [abandoned](http://www.xanthir.com/b4o00) from the CSS spec
* It is quite limiting in that you can't do media queries within them so you end
up with a mixture of @apply and @mixin

I still need to explore how to properly do fonts but for now there is
`helpers/mixins/fonts.css` to basically replicate what `@apply --font-16` was
doing


lewiscollard:
> It is quite limiting in that you can't do media queries within them so you end
up with a mixture of `@apply` and `@mixin`

I rather like the split between the two wherein we use `@apply` for simple property definitions and `@mixin` for more complicated ones (e.g. will need to work responsively).

dan-gamble:
I see where you're coming from but i'm not sure what falls in to `postcss-apply`. Also it was originally added as it was meant to be part of the CSS Spec.

For example fonts can't be `@apply`. Having the `button-base` can't be as you normally need to do `display: inline-block;`

Besides them 2 i don't think i used it much.

Do you have use cases where you use it frequently?

PR #216, opened 2017-10-26
[Resolves #203] Issue notify

### LINKED ISSUE #203: Look in to a circle "finished" hook that notifies certain users the push is now live.
Would need to check for `Fix #1` or `Fixes #1` in the commit message to see whether someone needs to be notified.

Could use the issue opener as the person to notify.

Not sure if Circle has hooks or is capable of doing this but could do it on the last step of `update.py` if it's not.

```
Knowing it and being given a notification saying you are now able to do "x" are 2 different things imo.

For example Fani is working on Quotient atm, she is doing content and filing issues. She gets a notification saying #26 has been closed, i don't think it would be in her head to check that in 10m time when she got the notification 10m ago.

That was my thought process behind the idea
```

danielsamuels:
When adding this to a project, you'll need to copy the environment variables from Bradfield as it contains the private key that's required for it to work.

You can do that by going to the repo settings in CircleCI, clicking "Environment Variables", and then "Import Variable(s)".

https://circleci.com/gh/onespacemedia/cser/edit#env-vars

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,8 @@ dependencies:
     - if psql demo -c ''; then dropdb demo; fi
     # Things required to create the project
     - pip install --upgrade pip setuptools cookiecutter python-coveralls
+    # For issue_notify
+    - pip install cryptography pygithub
     - npm install --global yarn
     - cookiecutter --no-input -f -o . --config-file tests/cookiecutter.yml .
     # Things required to test the project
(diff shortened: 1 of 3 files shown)


lewiscollard (approved):

dan-gamble:
Could the things needed for this be prompted for in the initial cookie cutter questions?

danielsamuels:
No, as it's only the single environment variable on CircleCI that's required. There isn't anything cookiecutter does related to this.

dan-gamble:
Fair enough!

danielsamuels:
Seems to be working well on a few projects.

Issue #195, opened 2017-10-13
Add CMS icons for common content types


lewiscollard:
Even on a relatively small project, the 'Add page' field has a bunch of identical icons. It would be nice if we could get some default icons for our most common `ContentBase` derivatives (and all of the project template apps).

![screen shot 2017-10-13 at 11 29 42](https://user-images.githubusercontent.com/4973097/31542411-c5e11cb2-b009-11e7-88e7-67845b16426d.png)

danielsamuels:
https://github.com/onespacemedia/cms-icons ?

lewiscollard:
How did I not think of that! I think I'm going to dry-run that on one of our current projects.

PR #217, opened 2017-10-26
[#195] Add icons for the common page content types.

lewiscollard:
So when adding a page, rather than a bunch of identical icons...

screen shot 2017-10-26 at 12 26 51

...we can now have this. :)

screen shot 2017-10-26 at 12 23 11

This also overrides the default CMS generic content icon and the 'Link' content type icon with some nicer ones that @fani04 made for us.

I'm happy to remove it and consider it an unrelated change, but this moved 'News' out of 'Syndication' and into 'Apps', where I think it makes more sense.


danielsamuels (approved):
I always hated syndication.

fani04:
That is great Lewis! It makes the CMS nicer!

dan-gamble (approved):

PR #219, opened 2017-11-02
[WIP] Edit bar

lewiscollard:
So here is a thing, experimental and possibly incomplete, to add an 'edit this thing' link. If you are on a detail page (i.e. `object` is in the template context), it'll give you an edit link if you are logged in as an admin. Otherwise, it'll give a link to edit the current page.

Testing this on TTP to work out any issues (see 102 over there), let's see how this works out. :)


danielsamuels:
I've got some thoughts on this, but I want to spend some time hacking around on it.

danielsamuels:
I've been playing with this a little on the TTP project and I've come up with the following iteration:



I think I prefer the bar at the top, as well as the ability to add another item alongside the existing edit link.

What do you think?

dan-gamble:
LGTM we might as well basically call ourself Django CMS now boys

danielsamuels (approved):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/site/templatetags/site.py
```diff
@@ -196,7 +198,67 @@ def md(value, inline=True):
...
+
+@library.global_function
+@library.render_with('edit_bar.html')
+@jinja2.contextfunction
+def edit_bar(context):
+    context = dict(context)
+    request = context['request']
+
+    # Don't show to non-admins, and don't pretend that /search/ is editable.
+    if not request.user.is_staff or request.path == '/search/':
```
dan-gamble, line 213: Should this take in to account page edit permissions as well?

dan-gamble (commented):
Has the edit bar had responsive done on it? Currently it looks like it doesn't even hide on mobile

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/site/templatetags/site.py
danielsamuels: It does on line 243.

dan-gamble: Whoops, misread what this line does 👍

PR #238, opened 2017-11-15
Base news app updates

dan-gamble:
Was a bit too much for images so made a video: https://www.youtube.com/watch?v=s2hZd3PEk-4

Tried to cut off a lot of the fat and slim it down
Made categories be a dumb model instead of a PageBase
Added default responsive styles for the list and detail
Added default call to action


dan-gamble:
Open to CC on this. Would like feedback on the default styles, i didn't do much (imo). I didn't do font sizing or the like or even the news cards. Just added something base to work with at the start of a project.

danielsamuels:
This looks great

jamesfoley (approved):

PR #247, opened 2018-01-25
Avoid opening mailto: links in a new window

danielsamuels:


lewiscollard:
What happens if their default mail handler is set to a webmail service (something I would not expect to be uncommon)?

Also out of curiosity, what happens if they have a program assigned to mailto links? A blank tab?

danielsamuels:
> What happens if their default mail handler is set to a webmail service (something I would not expect to be uncommon)?

Is that even possible?

> Also out of curiosity, what happens if they have a program assigned to mailto links? A blank tab?

Yeah, you basically get `about:blank` in a new tab.

danielsamuels:
Ok so it is possible and the answer is that you get gmail in your current tab.

lewiscollard:
In that case I think the more common case - i.e. probably ALL mobile devices, plus people on desktop with real mail clients installed - should be optimised. So this works for me.

lewiscollard (approved):

PR #248, opened 2018-02-08
[DO NOT MERGE] Fix CSS watch task on Linux

lewiscollard:


lewiscollard:
Scratch this for now - it might be breaking on OS X.

PR #265, opened 2018-05-02
Remove Safari-specific IFrame hack.

lewiscollard:
This was a work-around for a Vue issue, but as we now mount Vue selectively only where it's needed we don't need this anymore.


AidanCurrah (approved):
Seems good, approving

PR #266, opened 2018-05-03
Add a get_related_items helper function

lewiscollard:
This is fairly well-tested at this point (CSP, UK5G, BIPP) and it saves us re-implementing some version of it on every model. It doesn't make any assumptions about what model the related items use (you can easily, say, use it to get related events for a news article). Bonus: it updates `get_related_articles` on the news app.


AidanCurrah (approved):
Seems good, approving

PR #267, opened 2018-05-03
[WIP] Improve base for sections app 

lewiscollard:
* Add handling of transparent vs colour-background sections.
* Add standard section layout HTML (header/body/footer).
* Add a simple WYSIWYG text section to the standard sections.
* Add strcture for a standard hero section type


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/sections/models.py
```diff
@@ -39,17 +40,13 @@
     }),
     ('Text', {
         'sections': [
-            ('centered', {
-                'fields': ['background_colour', 'kicker', 'title', 'text', 'link_text', 'link_page', 'link_url'],
+            ('wysiwyg', {
+                'name': 'Rich text',
+                'fields': ['background_colour', 'kicker', 'title', 'content', 'link_text', 'link_page', 'link_url'],
                 'search': ['kicker', 'title', 'text'],
```
AidanCurrah, line 46: You've changed from text to content but not updated the search fields

lewiscollard: Good catch, thank you.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/sections/assets/css/section.css
```diff
@@ -0,0 +1,126 @@
...
+    padding-top: 3vr;
+    padding-bottom: @padding-top;
+  }
+
+  @media (--xlg) {
+    padding-top: 4vr;
+    padding-bottom: @padding-top;
+  }
+}
+
```
AidanCurrah, line 63: What are the benefits of using margins over padding? Couldn't we use padding to space all sections then we only need modifier classes  to change background-colour and colour.

lewiscollard: The reasoning behind this is that if you have two transparent (usually white-backgrounded) sections next to each other, the spacing between them will be enormous; this is not normally a problem for colour-backgrounded sections. Margins, on the other hand, collapse between adjacent items - the largest margin will take precedence.

PR #269, opened 2018-05-04
Added in-viewport checking for animation

joe-hastings:
Check if elements are currently in the viewport and add classes for animations, including staggering in multiple child elements


lewiscollard:
I think I am happy with this, but can you document its usage? We should be documenting all new features going forwards (and this is something I intend to do for everything that is already in the project template).

AidanCurrah (approved):
I can't see anything horribly wrong jumping out at me.

PR #275, opened 2018-05-29
Add a cookie bar

AidanCurrah:
The css is probably slightly off but hey


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/cookie-consent.css
```diff
@@ -0,0 +1,101 @@
...
+  height: 26px;
+  margin-left: 100px;
+
+  cursor: pointer;
+}
+
+.js-CookieOptIn,
+.js-CookieOptOut {
+  display: inline-block;
+}
```
lewiscollard: `.js-*`-namespaced classes should only be used for targeting things in JS, not CSS. Rename these classes to be under `.cc-Bar_*`.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base/_analytics.html
```diff
@@ -13,6 +13,17 @@
     })(window, document, 'script', '//www.google-analytics.com/analytics.js', 'ga');
 
     ga('create', '{{ settings.GOOGLE_ANALYTICS }}', 'auto');
+    // Only anonymize if not consented.
```
lewiscollard: Is there a reason we should not anonymise regardless of cookie consent?

lewiscollard: Also, when comments are in HTML, you should use template `{# comments #}` rather than `// comments` or `<!-- comments -->` - it is not useful to be in the rendered HTML.

lewiscollard: (I will update the style guide with this.)

AidanCurrah: I assumed the clients would want more accurate analytics if they could but I can change this to always anonymise IP

AidanCurrah:
[Add cookiebar to project template](https://trello.com/c/xuMmoptz/28-add-cookiebar-to-project-template)

lewiscollard (approved):

PR #277, opened 2018-05-30
RSS bug fixes.

jin-park-dev:
Add. process_rss method inside of ArticleFeedView to process cms content correctly for RSS.

url_from_path added to util to help getting correct base url.

URLs in <link> now returns correct absolute url now.

Manually setting content-length removed as it was giving wrong length. Django automatically adds content-length correctly.

Remove test relating to content-length.


jin-park-dev:
Tests added for process rss method.
Test added for utils method.
Fix iframe is now clickable rather than text.

lewiscollard (approved):

PR #278, opened 2018-05-31
Add pagination with concatenation of page list and styles.

AidanCurrah:
Implements the more recent pagination that displays pages like 1 ... 2 3 4 5 6 ... 10 along with some basic styles and the simple template.


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/pagination/pagination.html
```diff
@@ -0,0 +1,38 @@
...
+    {% if page_obj.has_previous() %}
+      <a href="{{ get_pagination_url(page_obj.previous_page_number()) }}" class="pgn-Pagination_Control pgn-Pagination_Control-prev" aria-label="Previous page">
+        {% include 'chevron-right.svg' %}
+      </a>
+    {% endif %}
+
+    {# using a separate element and aria-hidden because screen readers will
+    probably only care about previous/next #}
+    <div class="pgn-Pagination_Pages">
+
```
lewiscollard: Comment implies that this has `aria-hidden` on it, but it doesn't (it should)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/pagination/pagination.html
```diff
@@ -0,0 +1,38 @@
+{% if page_obj.has_other_pages() %}
+  <nav class="pgn-Pagination">
```
lewiscollard: An `aria-label` of `You are on page {{ page_obj.number }}` would be nice for screen readers.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/site/templatetags/site.py
```diff
@@ -209,3 +209,61 @@ def add_field_attributes(field, class_name, placeholder=True):
         'class': ' '.join((field.css_classes(), class_name)),
         'placeholder': field.label if placeholder else '',
     })
+
+
+@library.global_function
+@library.render_with('pagination/pagination.html')
+@jinja2.contextfunction
+def render_pagination(context, page_obj, offset=2, pagination_key=None):
+    '''Renders the pagination for the given page of items.'''
```
lewiscollard: As a docstring, this is incomplete; this comment should explain the truncation behaviour.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/pagination/pagination.html
```diff
@@ -0,0 +1,38 @@
...
+        {% if not page_num %}
+          <span class="pgn-Pagination_Page pgn-Pagination_Page-fake">
+            ...
+          </span>
+        {% elif page_num == page_obj.number %}
+          <span class="pgn-Pagination_Page pgn-Pagination_Page-current">
+            {{ page_num }}
+          </span>
+        {% else %}
+          <a class="pgn-Pagination_Page {% if page_num == page_obj.number %}pgn-Item_Action-current{% endif %}"
```
lewiscollard: This `if` condition will never be true, because it's sitting within the `else` clause of one that checks for this condition.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/pagination/pagination.html
```diff
@@ -0,0 +1,38 @@
...
+             href="{{ get_pagination_url(page_num) }}"
+             aria-current="{% if page_num == page_obj.number %}true{% else %}false{% endif %}">
```
lewiscollard: Same thing as above - this condition cannot be true. (And it makes little sense to have ARIA attributes on a block with `aria-hidden` - probably a hangover from the fact we were previously targeting this attribute rather than a class name as we do now.)

lewiscollard (changes requested):

lewiscollard (approved):

PR #281, opened 2018-06-05
Restructure JS to more consistent format

joe-hastings:
* main.js only has imports and initialisations of classes or functions, no logic
* Each folder has it's js in index.js for functions, classes get their own file with the filename same as the class name
* Change lazy image logic to be in its own module
* Simple function based calls (notices, viewport-animation, externalLinks etc.) sit in js/suitable-directory-name/index.js
* utils functions sit in `utils/index.js`
* shims sit in `compatibility`


lewiscollard (changes requested):
I would prefer that directories have `words-separated-by-spaces` like the individual JS files do, e.g. `lazy-images`, `focus-ring`, etc.

joe-hastings:
[PT: Come up with a agreed structure for JS files](https://trello.com/c/oE3ZaxH7/29-pt-come-up-with-a-agreed-structure-for-js-files)

lewiscollard (approved):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/external-links/index.js
```diff
@@ -0,0 +1,79 @@
+/*
+ External links (defined as all links pointing to a domain that is not the
+ same as the one from which the current document is being served)
+ will have their 'target' attribute set to '_blank'.
+
+ Social media sharing links will show in a popup.
+ */
```
daniel-clayton: Superfluous space here :triggered:

PR #286, opened 2018-06-11
Migrate the current moderation methods to use the CMS moderation

AidanCurrah:
Currently moderation is done via locally defined STATUS_CHOICES and string values. This updates this to use the CMS's moderation models and package.


joe-hastings (approved):
lgtm

PR #287, opened 2018-06-22
Add Rollbar JS exception logging

lewiscollard:


AidanCurrah (approved):

AidanCurrah:
The changes seem fine, but are we sure we want this on every project going forward? WNY is a little spammy with errors and I worry if we add this to every site we'll get flooded with small errors

lewiscollard:
I share that concern, but most of the errors come from very old browsers (early Android 4 in particular), e.g. `requestAnimationFrame is not defined`. While we don't support these browsers, I think it's reasonable to have a rule that we should never throw an exception (just `return` if certain things we require are not defined).

Basically, I think we should fix that at the coding standards level - I think it's too useful to know about JS errors. Though there's an open question of exactly how ancient we cater for in the 'never throw an exception' rule...

joe-hastings (approved):

PR #289, opened 2018-07-05
Implement section caching

lewiscollard:
Rendering sectioned pages tends to be one of the most expensive operations on a site. Caching sections can result in a substantial speed-up; in real-world testing this can knock a couple of hundred milliseconds or more from TTFB. I think this gives us much of the performance improvements from full-page caching, without any of the traditionally-troublesome edge cases of the latter. This is pretty thoroughly debugged at this point (five projects so far) and has never been a source of trouble.

It will never use the cache if someone is logged in as an admin user. This means that admins will see any changes they make immediately.

If a GET parameter `cache_bypass` is set to some non-empty value, then the cache will not be used, but it *will* be updated with the content of the section. This allows keeping the cache warm on certain important pages (like the homepage) by hitting the URL  every minute with `?cache_bypass=1` tacked on to the end.

The caching is intentionally stupid; it makes no attempt at cache invalidation when a section changes. I don't think this _can_ be done reliably, in principle. This is part of the reason behind the short expiry time.

Extra change: This disables automatic section template creation in `DEBUG` mode, which has been a plentiful source of "works on my machine" and threw `EEXIST` more often than not.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/sections/templatetags/sections.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/sections/templatetags/sections.py
@@ -1,8 +1,6 @@
-import errno
-
 import jinja2
 from django import template
-from django.template import TemplateDoesNotExist
+from django.core.cache import cache
 from django.template.loader import render_to_string
 from django_jinja import library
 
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/sections/templatetags/sections.py
```diff
@@ -12,26 +10,18 @@
...
+        return cached_content
 
-            os.system(f'touch {{ cookiecutter.package_name }}/apps/sections/templates/sections/types/{page_section.template["folder"]}/{page_section.template["file_name"]}')
+    context = dict(context)
+    context['section'] = page_section
 
-            return ''
-        else:
-            raise e
+    cached_content = render_to_string('sections/types/{}/{}'.format(
```
AidanCurrah, line 22: if the template/folder doesn't exist won't this throw an error? I know this shouldn't happen but it has recently, for example, on GPA with the text-centered section not having a template.

lewiscollard: Yep, and in my opinion, it _should_ throw an exception rather than failing silently.

AidanCurrah: Just making sure we don't mind this throwing an error

AidanCurrah (approved):

PR #292, opened 2018-07-30
Change mobile nav to be full height and fixed

joe-hastings:
Don't display the nav under the header instead make it full height and move the trigger on top. Also, make the burger menu use a CSS variable for its size rather than being hard-coded.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
@@ -101,7 +101,7 @@
 
 .nav-Mobile_Item,
 .nav-Mobile_DropdownItem {
- position: relative;
+  position: relative;
 
   flex-shrink: 0;
 
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
@@ -2,7 +2,7 @@
 |--------------------------------------------------------------------------
 | Mobile navigation
 |--------------------------------------------------------------------------\
-| Phone and tabet
+| Phone and tablet
 |
 | @namespace: nav-
 |
@@ -22,7 +22,7 @@
 
 .nav-Mobile {
   position: fixed;
-  top: var(--Mobile_NavHeight);
+  top: 0;
   right: 0;
   bottom: 0;
 
@@ -33,6 +33,8 @@
 
   font-size: 16px;
   font-weight: 600;
+
+  background-color: #fff;
 
   transform: translateX(100%);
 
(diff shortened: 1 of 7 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/navigation/navigation.js
```diff
@@ -20,34 +20,36 @@ export default class Navigation {
...
-      `.nav-Header_Items,
-      .nav-Header_Backdrop,
-      .nav-Header_Dropdown,
-      .nav-Mobile`
-    )
   }
 
+  /**
+   * If we're on mobile we want to track the selected states to control the sub navigation open
+   * / closed. aria-selected won't be used on lg++ so we programmatically add it just for sizes
```
lewiscollard: We shouldn't be using `aria-` attributes at all; the entire mobile nav is `aria-hidden="true"` so it makes no sense.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
```diff
@@ -1,3 +1,5 @@
+{% import 'components/macros/components.html' as components %}
```
lewiscollard: This should not be in the `components` app; it's site-wide and on every site so it shouldn't be in an app folder.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/navigation/navigation.js
```diff
@@ -20,34 +20,36 @@ export default class Navigation {
     this.setupListeners()
   }
 
+  /**
+   * Cache any dom elements we will use multiple times.
```
lewiscollard: DOM*

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
```diff
@@ -162,17 +185,19 @@
 */
 .nav-Mobile_Dropdown,
 .nav-Mobile_DropdownChild {
-  position: absolute;
-  top: 0;
+  position: fixed;
```
lewiscollard: I think I see a potential to break, though I haven't tested it: if you have a third level of navigation that has a transform on the parent (e.g. you're transitioning it in) won't `position: fixed` create a new stacking context which would then make the third level positioned `var(--Mobile_NavHeight)` from the height of its parent and *not* from the document?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/navigation/navigation.js
joe-hastings: It get's used for the styling of the current active items, as it's also `aria-selected` for desktop it kind of made sense to me to just leave it done like this for mobile too, I could replace with a modifier class for handling active items for mobile though if you'd prefer?

joe-hastings: Actually my bad was getting aria-current and aria-selected mixed up ignore the above 😁

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
joe-hastings: So should the macro go in `apps/site/templates/site/macros/site.html`?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/mobile.css
joe-hastings: Nice catch, tricky one to fix 🤔

lewiscollard (approved):

PR #294, opened 2018-08-03
Rationalise button mixins.

lewiscollard:
This is to make button names make more sense.

The old `-primary` and `-secondary` sometimes make sense if there are two button styles and there really is some kind of hierarchy rather than just different styles, but a) that is not universally the case, or even the case a majority of the time b) it sets a precedent to end up with `Button-tertiary` and `Button-quaternary`, &c.

Instead, let's not have opinions; make `Button` (previously did only the base styles) into `Button-base`, and make the button style simply `Button`. If we need multiple styles for a project, we'll name additional mixins by appearance, rather than role.


joe-hastings (changes requested):
Need to also change the classes in setting/base.py 357-364, probably ok to just have a single button here in the project template I reckon

lewiscollard:
@joe-hastings Good catch on `settings/base.py` - fixed.

joe-hastings (approved):

PR #296, opened 2018-08-14
Add 12-column grid overlay.

lewiscollard:
This adds a 12-column overlay for easier spotting of off-grid layout issues.

This is triggered from the edit bar. It's only visible to logged-in superusers.

It could probably do mobile better, but let's see how our mobile grid standardises.

Side effect: it standardises our media breakpoints in media-definitions.css to all use `>=` consistently; some were using `>`, some were using`>=`, and there needed to be one way to do it for breakpoint detection JS to work. I'm OK with either form, if it is used consistently.

screen shot 2018-08-14 at 12 57 49


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/base.html
```diff
@@ -151,6 +151,21 @@
...
     {% endblock %}
 
+   {% if request.user.is_superuser %}
+      <div class="dbg-Overlay">
+        <div class="dbg-Overlay_Breakpoint"></div>
+        <div class="dbg-Overlay_Inner">
+          <div class="dbg-Overlay_Row">
+            {% for _ in range(1, 13) %}
+             <div class="dbg-Overlay_Column">
+
```
joe-hastings: Some unneeded empty space bere

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/svg/fa-th.svg
```diff
@@ -0,0 +1 @@
+<svg class="svg-Image" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg"><path d="M512 1248v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm0-512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm640 512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm-640-1024v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm640 512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm640 512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm-640-1024v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm640 512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68zm0-512v192q0 40-28 68t-68 28h-320q-40 0-68-28t-28-68v-192q0-40 28-68t68-28h320q40 0 68 28t28 68z"/></svg>
```
joe-hastings: Is the class used for anything? Also, the filename for this SVG might be nice if it was descriptive i.e. grid.svg

joe-hastings (commented):
lgtm! Just a couple of small bits

joe-hastings:
Do you think it's necessary to have a way to set the grid size? GPA Global had a 14 col grid, so might need a way to adjust this from somewhere

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/svg/fa-th.svg
lewiscollard: The class sizes the SVG in CSS.

`fa-th` was to show that it comes from Font Awesome - same as the other ones - (this would be its name in HTML if we were using FA) but I can change that.

daniel-clayton:
@joe-hastings I think we should have a setting somewhere that defines this

daniel-clayton:
Then again we can always just change the magic number in range(1, n)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/svg/fa-th.svg
joe-hastings: Fair enough didn't spot that, I guess if we have other SVG's named in the same format as well then it makes sense for them all to be consistent, but not sure we really need to know where it came from - personally I'd just prefer a name that describes what it is

lewiscollard: ...but on second thoughts, you're totally right, I've re-organised these.

joe-hastings (approved):
I reckon this is good to go in, just changing the number of cols manually is alright really

PR #299, opened 2018-08-23
Add HasLinkMixin

AidanCurrah:
This mixin deals with the saving of the fields and also adds some nice helper functions for access the data. I've left sections alone and some of the links in the header.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/components/admin.py
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/components/admin.py
@@ -7,3 +7,8 @@ from .models import CallToAction
 @admin.register(CallToAction)
 class CallToActionAdmin(admin.ModelAdmin{% if cookiecutter.sections == 'yes' %}, UsedOnAdminMixin{% endif %}):
     list_display = ['title'{% if cookiecutter.sections == 'yes' %}, 'pages_used_on'{% endif %}]
+    fieldsets = (
+        (None, {
+            'fields': ['kicker', 'title', 'link_text', 'link_page', 'link_url']
+        }),
+    )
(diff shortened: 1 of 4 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/components/admin.py
```diff
@@ -7,3 +7,8 @@
 @admin.register(CallToAction)
 class CallToActionAdmin(admin.ModelAdmin{% if cookiecutter.sections == 'yes' %}, UsedOnAdminMixin{% endif %}):
     list_display = ['title'{% if cookiecutter.sections == 'yes' %}, 'pages_used_on'{% endif %}]
+    fieldsets = (
```
lewiscollard, line 10: Not that it makes much difference, but FYI you can this here instead:

```python
fields = ['kicker', 'title', 'link_text', 'link_page', 'link_url']
```

AidanCurrah: It's just my preference to have `fieldsets` in `admin.py`s and `fields` in `forms.py`s

lewiscollard: I'm OK with that. :)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/models.py
```diff
@@ -0,0 +1,49 @@
...
+    class Meta:
+        abstract = True
+
+    def clean(self):
+        if self.link_text and (not self.link_page and not self.link_url):
+            raise ValidationError({
+                'link_page': 'Please supply either a "link page" or "link URL"',
+            })
+
+    @property
```
joe-hastings, line 40: make this a `@cached_property`

joe-hastings (approved):

PR #300, opened 2018-09-28
Add columns Post CSS function for on grid width calculations

joe-hastings:
Add a Post CSS function that gives on grid width calculations for neater CSS so we don't need to have the long calc all over our code.

### DIFF
--- a/{{cookiecutter.repo_name}}/gulpfile.js/lib/postCssProcessors.js
+++ b/{{cookiecutter.repo_name}}/gulpfile.js/lib/postCssProcessors.js
@@ -1,12 +1,21 @@
-var path = require('path')
-
 module.exports = [
   require('postcss-easy-import'),
   require('postcss-sassy-mixins'),
   require('postcss-conditionals'),
   require('postcss-nested'),
   require('postcss-functions')({
-    glob: path.join(__dirname, '../../{{cookiecutter.package_name}}', 'assets', 'css', 'helpers', 'functions', '*.js')
+    functions: {
+      percentage: function (val1, val2) {
+        let number = val1 / val2 * 100
+
+        return number.toFixed(2) + '%'
+      },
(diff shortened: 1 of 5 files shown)


joe-hastings:
I suppose I could also go through and replace all the places widths that should be on grid are used to use this?

{{cookiecutter.repo_name}}/gulpfile.js/lib/postCssProcessors.js
```diff
@@ -1,12 +1,21 @@
...
   require('postcss-nested'),
   require('postcss-functions')({
-    glob: path.join(__dirname, '../../{{cookiecutter.package_name}}', 'assets', 'css', 'helpers', 'functions', '*.js')
+    functions: {
+      percentage: function (val1, val2) {
+        let number = val1 / val2 * 100
+
+        return number.toFixed(2) + '%'
+      },
+      columns: function (n, t, gutter) {
```
lewiscollard: Can you change `n` and `t` to full English words?

lewiscollard (changes requested):

joe-hastings (commented):
It's a bit odd how we use 8 columns at `--sm` and 12 at `--md`, we don't really tend to do this in new projects its always 12 all the way down to mobile, might be worth updating this?

lewiscollard:
Feel free to change that @joe-hastings, if you want. All but one of these are `width: columns(6, 8);`, which can be expressed as `columns(9, 12)` anyway.

joe-hastings:
> Feel free to change that @joe-hastings, if you want. All but one of these are `width: columns(6, 8);`, which can be expressed as `columns(9, 12)` anyway.

Cool think I will cause it's just easier to see whats going on at a quick glance, I can't be doing complex arithmetic like that on the spot ;)

lewiscollard (approved):

PR #302, opened 2018-12-17
Add frontend editing mixin

AidanCurrah:
Still needs:
- [ ] JS code cleanup
- [x] [Associated CMS small changes](https://github.com/onespacemedia/cms/pull/158)
- [ ] Allow users to view media library from HTMLField image selector
- [ ] Some styles/make it look good
- [ ] Edit mode toggle
- [ ] Un-hardcode post and get URLs
- [ ] Handle max_length of CharFields

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
@@ -62,6 +62,11 @@ export class SimpleEditable extends Editable {
     this.el.addEventListener('blur', () => {
       this.sendData()
     })
+
+    this.el.addEventListener('paste', e => {
+      e.preventDefault()
+      document.execCommand('inserttext', false, e.clipboardData.getData('text/plain'))
+    })
   }
 }
 
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
@@ -1,0 +1,107 @@
+import axios from 'axios'
+import tinymce from 'tinymce/tinymce'
+import renderWYSIWYG from './wysiwyg'
+
+class Editable {
+  constructor ({ el }) {
+    this.el = el
+    this.siteURL = '/frontend-edit/'
+
+    this.setupListeners()
+  }
+
+  setupListeners () {}
+
+  get editData () {
+    return {
+      app: this.el.dataset.app,
+      model: this.el.dataset.model,
+      pk: this.el.dataset.pk,
+      field: this.el.dataset.field,
+      value: this.el.innerHTML
+    }
+  }
+
+  sendData () {
+    const ajaxOptions = {
+      url: this.siteURL,
+      method: 'post',
+      data: this.editData,
+      xsrfCookieName: 'csrftoken',
+      xsrfHeaderName: 'X-CSRFToken',
+      headers: {
+        'X-Requested-With': 'XMLHttpRequest'
+      }
+    }
+
+    axios(ajaxOptions)
+    .then(response => {
+      return response
+    })
+    .catch(e => {
+      const { response } = e
+      const { status, data } = response
+
+      // form_invalid return
+      if (status === 400) {
+        return
+      }
+      /* eslint-disable no-console */
+      console.error(e)
+      /* eslint-enable no-console */
+    })
+  }
+}
+
+export class SimpleEditable extends Editable {
+  setupListeners () {
+    this.el.addEventListener('click', e => {
+      e.preventDefault()
+    })
+
+    this.el.addEventListener('blur', () => {
+      this.sendData()
+    })
+  }
+}
+
+export class WYSIWYGEditable extends Editable {
+  setupListeners () {
+    this.el.addEventListener('click', () => {
+      this.initialiseWYSIWYG()
+
+      this.saveButton.addEventListener('click', () => {
+        tinymce.remove(`#${this.el.getAttribute('id')}`)
+        this.saveButtonContainer.parentNode.removeChild(this.saveButtonContainer)
+
+        this.sendData()
+      })
+    })
+  }
+
+  initialiseWYSIWYG (){
+    renderWYSIWYG({selector: `#${this.el.getAttribute('id')}`})
+
+    this.saveButtonContainer = document.createElement('div')
+    this.saveButton = document.createElement('div')
+    this.saveButton.innerHTML = 'Save'
+    this.saveButton.classList.add('edt-SaveButton_Button')
+    this.saveButtonContainer.classList.add('edt-SaveButton_Container')
+    this.el.parentNode.appendChild(this.saveButtonContainer)
+    this.saveButtonContainer.appendChild(this.saveButton)
+  }
+}
+
+export class ImageEditable extends Editable {
+  setupListeners () {
+    this.el.addEventListener('click', () => {
+      if (!this.el.nextElementSibling || this.el.nextElementSibling.nodeName !== 'IFRAME') {
+        const iframe = document.createElement('iframe')
+        iframe.src = `${this.el.dataset.url}?file__iregex=.(png|gif|jpg|jpeg)$`
+        iframe.frameborder = '0'
+        iframe.classList.add('edt-Iframe')
+        this.el.parentNode.insertBefore(iframe, this.el.nextSibling)
+      }
+    })
+  }
+}
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
@@ -104,4 +104,14 @@ export class ImageEditable extends Editable {
       }
     })
   }
+
+  get editData () {
+    return {
+      app: this.el.dataset.app,
+      model: this.el.dataset.model,
+      pk: this.el.dataset.pk,
+      field: this.el.dataset.field,
+      value: this.el.dataset.value
+    }
+  }
 }
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
@@ -35,21 +35,21 @@ class Editable {
     }
 
     axios(ajaxOptions)
-    .then(response => {
-      return response
-    })
-    .catch(e => {
-      const { response } = e
-      const { status, data } = response
+      .then(response => {
+        return response
+      })
+      .catch(e => {
+        const { response } = e
+        const { status, data } = response
 
-      // form_invalid return
-      if (status === 400) {
-        return
-      }
-      /* eslint-disable no-console */
-      console.error(e)
-      /* eslint-enable no-console */
-    })
+        // form_invalid return
+        if (status === 400) {
+          return
+        }
+        /* eslint-disable no-console */
+        console.error(e)
+        /* eslint-enable no-console */
+      })
   }
 }
 
@@ -79,7 +79,7 @@ export class WYSIWYGEditable extends Editable {
     })
   }
 
-  initialiseWYSIWYG (){
+  initialiseWYSIWYG () {
     renderWYSIWYG({selector: `#${this.el.getAttribute('id')}`})
 
     this.saveButtonContainer = document.createElement('div')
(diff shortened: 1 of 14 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/wysiwyg.js
```diff
@@ -0,0 +1,95 @@
+import tinymce from 'tinymce/tinymce'
```
lewiscollard, line 1: I think there is probably a better answer than having a second copy of TinyMCE running on the frontend - we should be able to change TinyMCE settings in a single place (`settings/base.py`) rather than two.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/main.js
```diff
@@ -47,4 +48,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // Remove pre loading class, used for fixing some filer/animation issues on load
   removePreloadClass()
+
+  // Initialise front end editing
+  setupEditing()
```
lewiscollard: This should be wrapped in a `require.ensure` - there's no reason non-admin users of the site should have to load all of the editing code.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
```diff
@@ -33,6 +33,7 @@
 
     # Site URLs
     url(r'^assets/', include('django_lazy_image.urls', namespace='assets')),
+    url(r'^frontend-edit/$', FrontendEditView.as_view(), name='image_upload'),
```
lewiscollard: Why `image_upload`?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/models.py
```diff
@@ -0,0 +1,69 @@
...
+
+class EditableMixin:
+    def get_displayed_text_for_CharField(self, field_name, field_value):
+        # Need to handle max_length
+        return mark_safe(f'<span class="js-SimpleEditable" data-pk="{self.pk}" data-field="{field_name}" data-app="{self._meta.app_label}" data-model="{self._meta.object_name}" contentEditable>{field_value}</span>')
+
+    def get_displayed_text_for_TextField(self, field_name, field_value):
+        return mark_safe(f'<span class="js-SimpleEditable" data-pk="{self.pk}" data-field="{field_name}" data-app="{self._meta.app_label}" data-model="{self._meta.object_name}" contentEditable>{field_value}</span>')
+
+    def get_displayed_text_for_AutoField(self, field_name, field_value):
```
lewiscollard: Would you ever edit an AutoField?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
AidanCurrah: Ctrl + C is strong

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
...
+                if field and model._meta.get_field(field).get_internal_type() == 'ImageRefField':
+                    value = File.objects.get(pk=value)
+                if model and pk and field:
+                    candidate = model.objects.filter(pk=pk).first()
+                    setattr(candidate, field, value)
+                    candidate.save()
+                    return JsonResponse({'success': True})
+        return JsonResponse({'success': False})
+
+    def get(self, request):
```
lewiscollard, line 52: If this view's only purpose is to render the lazy image given a PK, move this into a different view.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
...
+        if request.user.is_superuser:
+            data = request.GET
+            image_pk = data.get('pk', False)
```
lewiscollard: Don't handle a "can't happen" case - use `data['pk']` because this should throw an exception if it's not present.

lewiscollard: (It also eliminates the branch below this line.)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
AidanCurrah: I meant it to be extendable so that in the future if we need to get more stuff from the server we can handle it here.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
         template_name = self.kwargs.get('slug', '_base')
 
         return ['frontend/{}.html'.format(template_name)]
+
+
+class FrontendEditView(View):
+    def post(self, request):
+        if request.user.is_superuser:
```
lewiscollard: The simplest branch (exit condition) should be evaluated first: do the permissions check and exit immediately if it fails. This makes it easier to read and thus reason about - as it is, nearly the whole of the code is contained in an if-branch.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
...
+            data = json.loads(request.body.decode('utf-8'))
+            app_name = data.get('app', False)
+            model_name = data.get('model', False)
```
lewiscollard: Once again, if you are never expecting these to _not_ be present, this should fail with an exception (e.g. a KeyError when accessing `data['app']`). It also eliminates a branch below.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
...
+            if app_name and model_name:
+                model = apps.get_model(app_label=app_name, model_name=model_name)
+                pk = data.get('pk', False)
```
lewiscollard: Same as before: is there any time this will be empty under reasonable circumstances? Let this fail loudly if not. (And it simplifies the `if` branches below.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/models.py
```diff
@@ -0,0 +1,69 @@
...
+from django.utils.functional import cached_property
+from django.utils.safestring import mark_safe
+from django_lazy_image.templatetags.lazy_image import lazy_image
+from threadlocals.threadlocals import get_current_user
+
+
+class EditableMixin:
+    def get_displayed_text_for_CharField(self, field_name, field_value):
+        # Need to handle max_length
+        return mark_safe(f'<span class="js-SimpleEditable" data-pk="{self.pk}" data-field="{field_name}" data-app="{self._meta.app_label}" data-model="{self._meta.object_name}" contentEditable>{field_value}</span>')
```
lewiscollard: This would make this entirely unsuitable to be used on user-supplied data; you've marked the entire string as safe which mean that if (for example) an article title contained user-supplied input (see any of our dashboard-type projects) there's a giant XSS here. Use [`format_html`](https://docs.djangoproject.com/en/2.1/ref/utils/#django.utils.html.format_html) for these.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/views.py
```diff
@@ -7,3 +18,36 @@ def get_template_names(self):
...
+                field = data.get('field', False)
+                value = data.get('value', False)
+                if field and not model._meta.get_field(field).get_internal_type() == 'HtmlField':
+                    value = strip_tags(re.sub('<br>', '\n', value))
```
lewiscollard: I agree with the sanitisation method here - as you've likely worked out this is one of the issues with `contentEditable` (pasting can give some pretty unpleasant HTML spam). While it's good to sanitise it on the server side, it's probably better if we do some client-side sanitisation too. You may want to force plain text only pasting in elements that should only contain plain text - [here](https://gist.github.com/cesarfigueroa/27d332384cad45217168) is an example to modify.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const WYSIWYGs = document.querySelectorAll('.js-WYSIWYGEditable')
+  for (const el of WYSIWYGs) {
+    new WYSIWYGEditable({ el })
+  }
+
+  const images = Array.from(document.querySelectorAll('.js-ImageEditable')).map(el => new ImageEditable({ el }))
+
+  window.addEventListener('message', (e) => {
+    if (e.origin === 'http://localhost:3000') {
```
lewiscollard: I am not sure of the purpose of this.

lewiscollard: Scratch that, I see it. It'd make more sense to check against `window.location.origin`. As elsewhere, the simpler, negative condition (event origin != the document's origin) should be checked against with an immediate exit rather than wrapping the entirely of the common case in the if clause.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/models.py
```diff
@@ -0,0 +1,98 @@
+from __future__ import print_function
```
lewiscollard: unnecessary, Python 2 is not going to be used for any future projects (and there should never be any `print`s in production anyway)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/editables/editable_base
```diff
@@ -0,0 +1,8 @@
+<span class="{% block class %}{% endblock %}"
```
lewiscollard: You should probably give this file an extension like the others.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/editable.js
```diff
@@ -0,0 +1,117 @@
...
+
+  setupListeners () {}
+
+  get editData () {
+    return {
+      app: this.el.dataset.app,
+      model: this.el.dataset.model,
+      pk: this.el.dataset.pk,
+      field: this.el.dataset.field,
+      value: this.el.innerHTML
```
lewiscollard, line 21: innerText might be better.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/js/frontend-editing/wysiwyg.js
lewiscollard: (The other option is that we have an extremely pared-down version of MCE for small edits - in which case, you can remove most of the plugins that have been loaded.)

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/utils/models.py
AidanCurrah: Dunno, I don't think it hurts to have the option?

lewiscollard: An AutoField is an auto-incrementing ID enforced by the database, as used in the `id` column on models and just about nowhere else - it is designed to be a permanent, immutable identifier. Even if it was ever rendered on the front-end I think it would be dangerous to allow it to be edited.

lewiscollard: An example of when editing this would be actively harmful, by the way. Say the highest number in an AutoField was 18, setting the next one to 19. If you *did* edit 18 to change it to 20, the database would probably happily let you do that. But as I recall (from having to fix this in the past from data imports gone bad) *this would not affect the column's sequence counter*. It would stay at 19, then get incremented to 20 on the next insertion, and on the *next* insertion you would encounter a uniqueness constraint violation as the database still thinks the next in the sequence is 20.

PR #303, opened 2019-03-06
Events app updates

daniel-clayton:

PR #304, opened 2019-03-07
Upgrade emails app

daniel-clayton:
- No longer renders emails using admin editable templates, instead users can fill in only specific fields of copy chosen by developers.

- Users must pick from a specific list of template references; you don't have to just know what these are or have a dev create them for you

- Should be impossible to 500 the site from the admin; if a email template doesn't exist for a specific reference then a new one is created with blank fields and the email is still sent. Developers still need to make sure they create the relevant templates in the backend, but this should avoid the issue of templates being forgotten between staging and production branches.

- Much simplified email template, using the retainer portal message as a base. Uses classes wherever possible and has some basic text styling.

There's a lot of ways to expand the functionality of this app, but I think this should be a decent v1 implementation that hopefully isn't too scary for clients to use - no jinja2 knowledge required!


daniel-clayton:
Failing builds fixed in #303

lewiscollard:
@daniel-clayton Would you merge develop into this to fix the build?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/contact/views.py
```diff
@@ -16,35 +14,17 @@ class ContactView(CreateView):
...
         context = {
+            'reply_to': form.cleaned_data['email'],
             'first_name': form.cleaned_data['first_name'],
             'last_name': form.cleaned_data['last_name'],
             'phone_number': form.cleaned_data['phone_number'],
             'reason_for_enquiry': form.cleaned_data['reason_for_enquiry'],
             'email': form.cleaned_data['email'],
             'job_title': form.cleaned_data['job_title'],
             'message': form.cleaned_data['message'],
-            'settings': settings,
```
lewiscollard, line 26: Is it not useful to have settings in the template context? e.g. `{{ settings.SITE_NAME }}`

daniel-clayton: Perhaps, but maybe it should be made available globally rather than just for the contact template?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/emails/templatetags/emails.py
```diff
@@ -13,3 +15,8 @@ def get_email_log_html(object_id):
 @register.assignment_tag()
 def get_email_template_html(object_id):
     return EmailTemplate.objects.get(pk=object_id).get_html_version()
+
+
+@library.global_function
+def email_url(path):
```
AidanCurrah: Is this needed as `path_to_url` is already a templatetag.

AidanCurrah (approved):

PR #308, opened 2019-03-11
Base styles and macros for lists and cards

daniel-clayton:


lewiscollard (approved):
I'm largely happy with this - it's running on two sites (therefore officially BugFree(TM)), plus these changes are something of a prerequisite for some other changes I want to make (see, e.g., aa2a93dec in the OSM site repo). The parts that make me nervous (the depth of macro/nesting) is undone easily enough if it annoys me.

PR #310, opened 2019-04-05
Upgrade to OAuth2

Kane1994:


lewiscollard:
Yeah, probably best if you don't commit our API keys. This is a cleanup job I did not want.

PR #316, opened 2019-04-25
Fix the admin bar to the bottom of the window rather than the top

lewiscollard:
We all seemed to agree that this would be better.

Bonus changes: simplify the CSS - don't set any properties on it at screen widths at which it is `display: none`. Make the SVGs work on IE11 since we officially support it (it needs explicit width and height - it won't fall back to the `viewBox` without them as other browsers will and it'll put it to some arbitrary size).

Tested on at least 2 sites so far, seems to work fine.

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/edit-bar.css
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/edit-bar.css
@@ -54,6 +54,7 @@
   cursor: pointer;
 
   .svg-Image {
+    width: 15px;
     height: 15px;
     margin-right: calc(var(--Grid_Inside) / 2);
   }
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/edit-bar.css
```diff
@@ -77,9 +82,9 @@
 */
 .pg-EditBar_FrameContainer {
   position: fixed;
-  top: 36px;
+  top: 0;
   right: 0;
-  bottom: 0;
+  bottom: 36px;
```
daniel-clayton, line 90: Where is 36 from?

lewiscollard: 24 (typical line height) + 6 px (0.25vr top padding) + 6px (0.25vr bottom padding). I'll 1) remove the 0.25 vr and replace it with '6px' (the styleguide says not to use very small fractional `vr`) 2) clarify this with a comment 3)  change the 40px used in `base/layout.css` which contradicts this.

AidanCurrah (approved):

PR #317, opened 2019-04-26
IE11 fixes for the edit bar

lewiscollard:
...since apparently people still use it in locked-down corporate environments and we do officially support it, for now.

Fixes:

1) It no longer uses ES6 syntax that IE11 does not support like fat arrows
2) This removes an errant `use strict` that IE11 (probably correctly) complains about.


joe-hastings (changes requested):
It's working in IE11 but there's some CSS issues too:
ie11 edit bar

lewiscollard:
@joe-hastings Thanks for spotting the omission - try it with the latest CSS :)

joe-hastings (approved):

PR #321, opened 2019-05-03
Accessibility: Make it possible to tab to dropdown menus.

lewiscollard:
As titled. As well as showing dropdowns on hover, we can also do them when the element receives focus (via tab completion). While it's actually the link getting focus, there's [this (MDN)](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within):

> The `:focus-within` CSS pseudo-class represents an element that has received focus or contains an element that has received focus.

In testing on a current project, let's see how that goes before merging. (Edit: Tested, works.)

### DIFF
--- a/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/desktop.css
+++ b/{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/assets/css/globals/navigation/desktop.css
@@ -202,6 +202,12 @@
     opacity: 1;
     visibility: visible;
   }
+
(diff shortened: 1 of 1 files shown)


joe-hastings (changes requested):
Need to add the same focus-within selector for dropdown children so it works for multiple levels of child pages, add:
```
  .nav-Header_Item-hasDropdown:focus-within > & {
    opacity: 1;
    visibility: visible;
  }
```
inserted at line 204

lewiscollard:
@joe-hastings Fixed!

joe-hastings (approved):

PR #323, opened 2019-05-09
Restore Spread mixin.

lewiscollard:
Since this almost invariably finds its way back into every project.


joe-hastings (commented):
I've seen in some versions of this mixin 
```
  /* special case for iframes */
  width: 100%;
  height: 100%;
```
I'm not sure if it's actually needed or not, not sure who/when it was added as the mixin gets copied around from project to project.

joe-hastings:
> I've seen in some versions of this mixin
> 
> ```
>   /* special case for iframes */
>   width: 100%;
>   height: 100%;
> ```
> 
> I'm not sure if it's actually needed or not, not sure who/when it was added as the mixin gets copied around from project to project.

Thinking about it I think my preference is to not have the extra special case in the mixin, if this is ever needed it can just be included as and when it's needed

joe-hastings (approved):

lewiscollard:
As I recall I put that special case in originally. I think it was a cross-browser-BS issue - probably when iframes have explicit width/height attributes but I don't entirely recall. :/

joe-hastings:
I reckon if we come across needing to add it again then we'll add it to the mixin but for now I like the feeling that all the spread mixin does is the position and t r b l positions

PR #326, opened 2019-05-17
Add smooth_gradient PostCSS function that allows for easing function based gradients

AidanCurrah:
I am unsure of whether to raise an error is an non-easing function is entered or simply fallback to a default, say `ease`.


{{cookiecutter.repo_name}}/gulpfile.js/lib/postCssProcessors.js
```diff
@@ -13,6 +13,51 @@ module.exports = [
         gutter = gutter ? `${gutter}px` : 'var(--Grid_Gutter)'
 
         return `calc((${columns} / ${totalColumns} * 100%) - (${gutter} * 2) * (${totalColumns} - ${columns}) / ${totalColumns})`
+      },
+      smooth_gradient: function (direction, startingColor, easingFunction, finalColor = 'transparent', stops = 5, colorMode = '') {
```
lewiscollard: For consistency with how functions are always named in CSS, can this be called `smooth-gradient` rather than `smooth_gradient`? I hope postcss-functions allows this.

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/gulpfile.js/lib/postCssProcessors.js
lewiscollard: (That should be "things that sort of look like function calls", like `linear-gradient`.)

AidanCurrah: Nope. It can't have a `-` in it since it's a javascript function and it can't be camelCase since PostCSS didn't like that in the css files

AidanCurrah: It is annoying I know

lewiscollard: What happens if you use a quoted string as the object key?

lewiscollard:
Happy with this now. I've changed it to be `smooth-gradient` (you never see underscores in CSS), to require an end colour, and to have a default for `easingFunction`. Those were only nitpicks really. Good work!

lewiscollard (approved):

PR #333, opened 2019-07-24
Remove browser-update.org.

lewiscollard:
First, our browser support is already extremely generous; anyone who is seeing a broken site is probably seeing an entirely broken Web already.

Second, It makes me uncomfortable loading third-party JS without any integrity checks.

Finally, we've killed it on three projects so far and we've had no complaints.


daniel-clayton:
I agree

daniel-clayton (approved):

PR #335, opened 2019-09-04
Remove auto-generated summaries for news articles.

lewiscollard:
My opinion is that humans should be writing summaries.

Summaries that are generated from the first N words of the article are a lottery of usefulness; if they give a usable overview of the article it is only by luck. If summaries are required for layout purposes, the summary field should be required. If they are not required for layout purposes, then one does not need to be generated.


daniel-clayton:
It's useful for when you have a site with lots of imported news articles though.

lewiscollard:
I disagree - that just gives you a site that has lots of news articles with nonsensical summaries.

daniel-clayton:
But it's useful filler when you might have a few bespoke news articles created for the new site and 200 imported ones. As you say, it might be important for layout purposes but no one is going to go through and fill out the summaries for old articles.

PR #336, opened 2019-10-09
Add a test page for WYSIWYG styles.

lewiscollard:
This adds a template for testing every possible WYSIWYG style (it can be viewed through `/frontend/wysiwyg/` when `settings.DEBUG == True`.

Bonus changes:
* Cells of the first row of a table are not treated as a header row - one should use an actual header row if that is required. 
* Default styles are now added for _unclassed_ h[2-6].
* A never-used `wys-Title-xlarge` option is removed from the default settings.
* The `paste` plugin is restored (required for paste as text to work).


daniel-clayton (commented):
image
Does this image work for you?

daniel-clayton (commented):
image
I feel like we should just make `hr`s `margin: 1vr 0`

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/frontend/wysiwyg.html
```diff
@@ -0,0 +1,82 @@
+{% extends 'frontend/_base.html' %}
+
+{% block main %}
+  <div class="wys-WYSIWYG" style="max-width: 800px">
+    <p class="wys-Standfirst">Standfirst lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>
```
daniel-clayton: Can we include the class `wys-Standfirst` in `wysiwyg.css` even if it just has `color: var(--Colour_Something)`

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/base.py
```diff
@@ -404,11 +404,6 @@
                     'selector': 'h1,h2,h3,h4,h5,h6',
                     'classes': 'wys-Title-large'
                 },
-                {
-                    'title': 'Extra large',
-                    'selector': 'h1,h2,h3,h4,h5,h6',
-                    'classes': 'wys-Title-extraLarge'
-                }
             ]
```
daniel-clayton, line 407: Missing the option to select `wys-Standfirst`

lewiscollard:
@daniel-clayton ah of course that image worked on the particular project I copied it from... :/ There's no reasonable way to test that, so I've removed the image from the test (it can't really go wrong anyway).

`<hr>`s in WYSIWYGs are now `margin: 1vr 0`.

daniel-clayton (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/settings/base.py
lewiscollard: My provisional solution to this is to remove the standfirst test in the frontend template, as not all sites use a standfirst style.

daniel-clayton:
Do you want to squash and merge, since the build only failed because of the pytest thing?

daniel-clayton (approved):

AidanCurrah (approved):

PR #340, opened 2019-10-29
Add password reset functionality

daniel-clayton:
- Makes the emails app a mandatory app


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/emails/templates/emails/password-reset.html
```diff
@@ -0,0 +1,22 @@
...
+  <p>If you want to reset your password <a href="{{ protocol }}://{{ domain }}{{ url('password_reset_confirm', uidb64=uid, token=token) }}">please click here</a>.</p>
+
+  <p>Please note that this reset link will expire in 3 days.</p>
+
+  <p>Common login problems:</p>
+
+  <ol>
+    <li>Only use the email address that is registered with {{ settings.SITE_NAME }} (that will be the one that this email has been sent to).</li>
+    <li>Double check the login details you are entering. Be careful when typing your email address and password as sometimes it's easy to mistype them especially on tablets and phones.</li>
+    <li>Close and restart your browser, and ideally clear your cache, cookies and history, before attempting to log in with your new password.</li>
```
lewiscollard: You should never need to do this. Clearing cookies is a last resort, and the browser history makes no difference.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/users/templates/users/reset/password_reset_complete.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'users/reset/password_reset_base.html' %}
+
+{% block content %}
+  <h1 class="ath-Form_Heading">Password reset successful.</h1>
+
+  <p class="ath-Form_Message">Your password has been reset. You can now <a href="/admin/login/">log in</a> again.</p>
```
lewiscollard: Don't assume where the admin login form lives - `{% url %}` it.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
```diff
@@ -25,6 +26,15 @@
     url(r'^admin/', include(admin.site.urls)),
     url(r'^admin/', include('social_django.urls', namespace='social')),
     {% if cookiecutter.sections == 'no' %}# {% endif %}url(r'^admin/pages/page/sections.js$', sections_js, name='admin_sections_js'),
+    url(r'^admin/reset-password/$', reset_views.ProjectPasswordResetView.as_view(), name='password_reset'),
```
lewiscollard: Break these out into a urlconf in the `users` app, IMO.

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/static/admin/auth/login.css
```diff
@@ -0,0 +1,319 @@
+/*
```
lewiscollard: Was this file included accidentally?

lewiscollard (changes requested):

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/static/admin/auth/login.css
daniel-clayton: It has the form styles that are used

lewiscollard: OK, some parts are extremely similar to the `auth/login.css` in `osm-jet`. Is there a case for moving it there?

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/urls.py
daniel-clayton: Would it not be a bit strange that some `/admin/` urls go through the users app and some don't?

daniel-clayton: Then again so do the social urls

lewiscollard: I don't think it is any aesthetically weirder, and having a separate urlconf would keep app-specific URL routes within the app.

daniel-clayton: kdoke, but I'm not going to use a namespace because it would involve doing a bunch of overrides

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/emails/templates/emails/password-reset.html
```diff
@@ -0,0 +1,22 @@
+{% extends 'emails/base.html' %}
```
lewiscollard: Django convention is to use underscores to separate words in template names. This pull request uses dashes inconsistently, too.

lewiscollard: ^^ I think you may have pushed up the fix for this one about five seconds before I wrote this comment :)

lewiscollard (approved):

PR #341, opened 2020-01-24
Update the project template to be Django 2.2 compatible

AidanCurrah:


{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/apps/redirects/admin.py
```diff
@@ -75,14 +75,14 @@ def get_list_display(self, request):
...
 
         if not getattr(settings, 'REDIRECTS_ENABLE_REGEX', False):
             kwargs.update({
                 'exclude': ['regular_expression', 'test_path']
             })
-        form = super().get_form(request, obj, **kwargs)
+        form = super().get_form(request, obj=obj, change=change, **kwargs)
         return form
 
     def test_redirect(self, obj):
```
joe-hastings, line 89: need to `mark_safe` the return for `test_redirect` or it renders the html:

`from django.utils.safestring import mark_safe`

`return mark_safe('<a target="_blank" href="{}">Test</a>'.format(escape(url)))`

{{cookiecutter.repo_name}}/{{cookiecutter.package_name}}/templates/includes/featured_card.html
```diff
@@ -11,7 +11,7 @@ <h2 class="art-FeaturedCard_Title">{{ object.title }}</h2>
 
   {% if object.card_image or object.image %}
     <div class="art-FeaturedCard_Image">
-      {{ render_lazy_image(object.card_image or object.image, width=843, height=480, blur=False, show_small_image=False) }}
+      {{ _osm_image(object.card_image or object.image, width=843, height=480, blur=False, show_small_image=False) }}
```
joe-hastings: typo, should be `render_osm_lazy_image`

joe-hastings (changes requested):
Looks fine apart from the couple of comments

joe-hastings (approved):

PR #342, opened 2020-01-24
Update frontend build system to use Webpack4

AidanCurrah:
This should make it easier to extend our system in the future. It also has the side benefit of no longer having a max depth for linting so our sections CSS files should get linted now


joe-hastings (approved):
All looks good to me

